PR #12, opened 2021-05-11
Add annotated i/o for real-time clients

weefuzzy:
Quick & dirty. 
 * part of fix for flucoma/flucoma-max#2
 * Adds the ability to associate text with Client inputs and outputs to support inlet/outlet assistance in Max. This anticipates a planned change to the way we declare I/O in the Clients after we finishing refactoring them. 
 * Text is only needed and provided for the RT clients at this point because the labelling for all buffer processor and model object inlets / outlets will be identical across objects in Max.


tremblap:
Still runs all the UTs on SC, will use it  in my current workflow a bit to check for stability.

PR #17, opened 2021-06-25
Include offsets in buffer range checks for actual checkingness

weefuzzy:
Clearly people are making heavy use of our frame and channel offsets 😆 

We seem to have made it this far without me factoring these into the range checking used in the NRT wrapper and elsewhere, as just discovered by @g-roma


weefuzzy:
This doesn't seem to have been merged yet. I think we meant to?

PR #26, opened 2021-09-08
Add option to drop 0th MFCC

weefuzzy:
Adds a new param to MFCC to drop the 0th coefficient. Will automatically show up in Max, and if we like it we can add to the SC class.

### DIFF
--- a/include/clients/rt/MFCCClient.hpp
+++ b/include/clients/rt/MFCCClient.hpp
@@ -41,7 +41,7 @@ constexpr auto MFCCParams = defineParameters(
               UpperLimit<kNBands, kMaxNCoefs>()),
     LongParam("numBands", "Number of Bands", 40, Min(2),
               FrameSizeUpperLimit<kFFT>(), LowerLimit<kNCoefs>()),
-    EnumParam("dropZero","Drop zeroth Coefficient",0, "No","Yes"),
+    EnumParam("dropZero","Drop Zeroth Coefficient",0, "No","Yes"),
     FloatParam("minFreq", "Low Frequency Bound", 20, Min(0)),
     FloatParam("maxFreq", "High Frequency Bound", 20000, Min(0)),
     LongParam<Fixed<true>>("maxNumCoeffs", "Maximum Number of Coefficients", 40,
--- a/include/clients/rt/MFCCClient.hpp
+++ b/include/clients/rt/MFCCClient.hpp
@@ -41,7 +41,7 @@ constexpr auto MFCCParams = defineParameters(
               UpperLimit<kNBands, kMaxNCoefs>()),
     LongParam("numBands", "Number of Bands", 40, Min(2),
               FrameSizeUpperLimit<kFFT>(), LowerLimit<kNCoefs>()),
-    EnumParam("dropZero","Drop Zeroth Coefficient",0, "No","Yes"),
+    LongParam("startCoeff","Output Coefficient Offset",0),
     FloatParam("minFreq", "Low Frequency Bound", 20, Min(0)),
     FloatParam("maxFreq", "High Frequency Bound", 20000, Min(0)),
     LongParam<Fixed<true>>("maxNumCoeffs", "Maximum Number of Coefficients", 40,
--- a/include/clients/rt/MFCCClient.hpp
+++ b/include/clients/rt/MFCCClient.hpp
@@ -28,6 +28,7 @@ namespace mfcc {
 enum MFCCParamIndex {
   kNCoefs,
   kNBands,
+  kDrop0,
   kMinFreq,
   kMaxFreq,
   kMaxNCoefs,
@@ -40,6 +41,7 @@ constexpr auto MFCCParams = defineParameters(
               UpperLimit<kNBands, kMaxNCoefs>()),
     LongParam("numBands", "Number of Bands", 40, Min(2),
               FrameSizeUpperLimit<kFFT>(), LowerLimit<kNCoefs>()),
+    EnumParam("dropZero","Drop zeroth Coefficient",0, "No","Yes"),
     FloatParam("minFreq", "Low Frequency Bound", 20, Min(0)),
     FloatParam("maxFreq", "High Frequency Bound", 20000, Min(0)),
     LongParam<Fixed<true>>("maxNumCoeffs", "Maximum Number of Coefficients", 40,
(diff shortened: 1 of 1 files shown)


jamesb93:
I think that this change has a lot of value for reducing repetitive patching. That said, I don't like @drop0 and would prefer @dropzero.

weefuzzy:
> I think that this change has a lot of value for reducing repetitive patching. That said, I don't like @Drop0 and would prefer @dropZero.

It is `@dropzero` ❤️

jamesb93:
Then you have my blessing ⛪ 🙏

tremblap:
Things are never as simple as they seem. The current implementation works but has one issue of interface.

If I ask for 13 MFCC and to drop0, I was expecting to get 12 coeffs - computing 13, getting 12, from 1 to 12, droping 0 in 0-counting

now I get 13, so 1-13 and therefore computing 14 and droping 1.

I vote for the first behaviour which is more in line with what is happening. also it helps with the problematic situation:
[fluid.mfcc~ 13 @numbands 13 @dropzero 1]

in this case there are only 13 bands to compute the 14 coeffs to output 13. I'm surprise I don't get a warning that this is not working under the hood...

jamesb93:
I agree, although I think that none of us disagree and the implementation is just doing Extra Counting 🔢 . The dropping should be "post-fader", rather than "pre-fader". If I ask for 13, then dropZero === 1 I should get 12 coeffs at the end of that process.

weefuzzy:
That was deliberate, because my intuition about what constituted sensible behaviour was the opposite to both of yours 😄 ISTM, that if I ask for 13 then the least surprising behaviour is that is what I get, irrespective of whether I'm counting from 0 or 1. Otherwise it's a magical implementation detail that users have to first know about and then remember (every time).

jamesb93:
I sort of agree that a "what you specify is what you get" is sane and dependable but either way there is some level of user have to look up the behaviour. My view is that it depends on the user's perspective of what they think they are setting in the flags. If I set `@numcoeffs` to 13 but I remove the 0th coeff then a result with 13 coeffs seems *odd*. Isn't it also fundamentally a different calculation? i.e setting @numcoeffs to 13 and then computing a 14 coeff MFCC in order to drop 0 might skew results differently or be surprising too.

tedmoore:
To me, "dropZero" sounds like a post-calculation action. As in, "Do the calculations as I've specified, then once you have that zero thing...drop it". Which is to say that I think if a user asks for 13 and flags dropZero, the object should output 12. I would be confused if I asked for 13 and said "dropZero" and got 13. Like, "Wait, I thought I said to drop the zeroth one...?"

weefuzzy:
> Isn't it also fundamentally a different calculation? i.e setting @numcoeffs to 13 and then computing a 14 coeff MFCC in order to drop 0 might skew results differently or be surprising too.

No. Changing the underlying number of Mel bands would be a fundamentally different calculation, but here you're literally just deciding which cepstral coefficients to keep. The normal effect of doing the DCT would result in as many output coefficients as the size of the input. I don't know if G's code does anything special or different there, but in general the liftering is just chucking stuff away

jamesb93:
> > Isn't it also fundamentally a different calculation? i.e setting @numcoeffs to 13 and then computing a 14 coeff MFCC in order to drop 0 might skew results differently or be surprising too.
> 
> No. Changing the underlying number of Mel bands would be a fundamentally different calculation, but here you're literally just deciding which cepstral coefficients to keep. The normal effect of doing the DCT would result in as many output coefficients as the size of the input. I don't know if G's code does anything special or different there, but in general the liftering is just chucking stuff away

Gotchya. That makes sense when you know the algorithm a bit better than I do then 😆

weefuzzy:
> To me, "dropZero" sounds like a post-calculation action.

Maybe it's a naming problem then. I feel pretty strongly that if I say `[fluid.mfcc~ 13 ...]` then future me will *always* be confused by subsequent code that assumes 12 thing. We could make it a more general `startFrom` or something (which opens the possibility of easier experimentation with higher coefficients).

tedmoore:
I do like `startFrom`.

tedmoore:
Just a comment on the term, `startFrom`, perhaps `startCoeff` would be better as it's more specific and is more in line with the terminology in other objects, such as `startChan` and `startFrame`

jamesb93:
> Just a comment on the term, `startFrom`, perhaps `startCoeff` would be better as it's more specific and is more in line with the terminology in other objects, such as `startChan` and `startFrame`

This begins to make me feel more warm and fuzzy inside 🔥  Also quite powerful if you just want the top half of your coefficients. Again, encroaching on another objects role I think and is a slippery slope to every object having their own input and output @start/end

weefuzzy:
> perhaps `startCoeff` would be better

Agreed 

> we have bufselect~ or bufcompose~ to deal with data manipulation

But the motivation here is to remove the need to do this very often-needed extra work, viz. given that much of the time coeff 0 is surplus to requirements, it's a pain to have to slice / select (and less easy to teach)

jamesb93:
> But the motivation here is to remove the need to do this very often-needed extra work, viz. given that much of the time coeff 0 is surplus to requirements, it's a pain to have to slice / select (and less easy to teach)

Without trying to sound belligerent and rubber-bandy, I don't see the issue with `@dropzero` then as a name as its specific to a repetitive and oft-needed process. Easy to explain to a newcomer that you might think about dropping your 0th coeff as its a representation of power, or an offset to the spectrum. Whether or not it counts from 0-1 is to me one way in my head, (a post-calculation process) but I recognise that its really a product of my engagement with that on a personal level and am totally open to the other ways of reasoning about it.

tremblap:
So for me, the most potent solution so far is @fromCoeff for 4 reasons you all mentionned

1. your ask for 13 you get 13 (and not 13 or 12 depending on something else)
2. you don't get what you don't ask (ask for 13 and drop, expect 12 if you understand it as post yet get 13 aka another coef)
3. more interestingly, you can skip the first 5 if you want - much more experimental interface
4. default of startCoeff at 0 is backward compatible

I find this is the best compromise.

weefuzzy:
Cool. I'll update this PR accordingly (later, I'm in the weeds with something else just now)

tedmoore:
Greetings @weefuzzy, I'm hoping to use this PR this week. Just checking on the status--if there's a particular day I should aim for? Thanks.

weefuzzy:
I can rename the attribute very quickly, if you just want to get going. Doing `startCoeff` > 1 requires a bit more work.

tedmoore:
@weefuzzy thanks!

jamesb93:
Just tested building and playing. This works for me and I think the interface is fine. I believe it will need very clear docs and clear explanation in the reference.

jamesb93 (approved):

weefuzzy:
Cool, cheers. Not quite finished yet (startCoeff > 1 won't work at all I don't think)

jamesb93:
I assumed as much, but for startCoeff <= 1 seems to play ball :)

tremblap:
Hello! As I was modding the SC class def, I noticed a discrepency:

fromCoeff should be startCoeff

I can change it in Max too if we agree.

tremblap:
Also, as I am bringing the SC class, the order is messy now:

`startFrame = 0, numFrames = -1, startChan = 0, numChans = -1, features, numCoeffs = 13, numBands = 40, startCoeff = 0,`

but maybe we can live with it in that order since 1) there is autocomplete and 2) startCoeff is definitely niche

comments?

tedmoore:
I am not bothered by this order

tremblap:
ted had proposed it, order stays the same, testing now to check and then updating 3 repos. I'll add Min and Max check to the client with a note too

PR #29, opened 2021-09-15
update HISSTools to tip

jamesb93:
This will allow fat binaries to be built on ARM/m1 machines by pushing the HISSTools dependency to the tip of its main branch.


weefuzzy:
Builds fine here. Will require proper testing though as there are extensive looking changes that affect many of our things. Will run against the SC unit tests on mac, win and linux

jamesb93:
I'm actually going to close this for now. We can reopen when necessary but IMO it clutters the working state of the repository and is not that essential when releasing although IMO we need to bump `main` in order to make automatic pulling of deps build fat binaries at some point.

PR #34, opened 2021-09-20
Fix crash due to incorrect removal of dead param listeners for shared objects

weefuzzy:
A slightly embarrassing bug for shared objects that caused crashes when named-shared objects (only named-shared objects were being deleted). Unlikely to be seen often in SC, but could have caused the odd crash in Max with `dataset~/labelset~` when reinstantiating boxes. Due to there now being many more named-shared types on my branch-in-progress, I ran into this during testing. 

For future ref, one needs to remember (d'oh) that `std::remove_if` always needs to be used in conjunction with an actual `erase` operation on the container in question: `remove` only shuffles things about, it doesn't actually do the erasure. If you forget this, then one's container still has dead things littering it.

### DIFF
--- a/include/clients/nrt/FluidSharedInstanceAdaptor.hpp
+++ b/include/clients/nrt/FluidSharedInstanceAdaptor.hpp
@@ -203,8 +203,10 @@ public:
   void removeListener(void* key)
   {
     auto& listeners = mParams->listeners[N];
-    std::remove_if(listeners.begin(), listeners.end(),
-                   [&key](ListenerEntry& e) { return e.second == key; });
+    listeners.erase(
+        std::remove_if(listeners.begin(), listeners.end(),
+                       [&key](ListenerEntry& e) { return e.second == key; }),
+        listeners.end());
   }
 
 private:


jamesb93:
Is there a way to recreate or is it a bit slippery?

weefuzzy:
It was very easy on my feature branch 🤣 

I think it might be slippery against `main` because `dataset / labelset` don't have any `@` attributes, and the easiest way to trigger it is with re-making a box with attribute values being set in-box. 

I guess, if you really want to verify that the crash was a thing, for the purposes of experimentation 
* you could add a dummy param to dataset and labelset and build by adding 
  ```c++
  LongParam("dum","Dummy",0)`
  ```
  to the `defineParameters` block (*after* name param, that's important) 
* set this in a box `[fluid.dataset @dum 1]`
* remake the box, perhaps with a different value for `dum` (you might need a 2nd instance with the same name in the patch, not sure)

jamesb93:
Can't crash on my end so assume its fixed?

jamesb93 (approved):

tremblap:
it works - I don't know how to test but everything works.

PR #35, opened 2021-09-20
Enable naming / sharing for all RT queryable model objects

weefuzzy:
This addresses a long-standing wish to be able to have sharing for those model clients that support `transformPoint / predictPoint` (i.e. that set that has corresponding UGen query objects in SC), so that e.g. `fluid.kdtree~` in Max / PD can be given a name and queried in another part of the patch without spaghetti or send / receive.  

There are supporting PRs in the Max, SC and docs repos too. PD needs to catch up (again).  

This approach to the problem was enabled by (finally) implementing control rate input for a rolling stats object. The approach taken here to enabling naming-sharing for these clients is to factor out the 'RT' query machinery (which is only useful in SC as it stands) into separate clients, so that the base client can then take advantage of the sharing framework, which only works for NRT objects (because reasons). E.g., for KDTreeClient there is an accompanying KDTreeQuery that uses an instance of KDTreeClient's shared type. 

So, there are a few points of possible controversy, but also some benefits: 
* 👍  Obvisouly the fact of being able do to this improves UX in Max/PD substantially. 
*  👍  Creating separate clients like this means a *reduction in magic*: it was never clear for the uninitiated why some ModelObjects had audio inputs and a RT process method. This way is more explicit, IMO, and also solves the resulting weirdness of having redundant in/out buffer parameters 
*  👎 It does, however, introduce further boilerplate to the client headers. Longer term work to solve that more generally continues 
* 👎  The possible forthcoming fix for ModelObject clients being able to push parameter changes back to the CCE after `load` gets broken by this, as it relied on a implementation detail due to the affected objects being RT. I don't see this as a huge drawback and would prefer to implement a principled way to do this instead, via `FluidContext`. 

AFAIK, nothing else gets broken by this (although the Dafx graph-loop objects *might* be?). The user facing behaviour in SC should be identical to what it was. In Max, we have names and no more redundant attributes.


jamesb93:
This builds for me and I can't get it to break. Tried a few things like wonky names (\/\/\/\\/\\\ for example) as well as giving the same name to lots of objects (a fluid.normalize~, a fluid.mlpclassifer~, a fluid.umap~) which didn't produce nefarious results. This is new, so hard to evaluate against existing patches for breakage but I'm personally satisfied with the rudimentary testing I did. I don't see a problem in the implementation either, however, I'll need to work on putting this in help files sharpish.

jamesb93 (approved):

PR #60, opened 2021-10-30
MFCCs: Ensure DCT reinits to correct size on `reset`

### LINKED ISSUE #59: bufmfcc spits out DC on the last coeff when @startcoeff = 1
as shown here
https://discourse.flucoma.org/t/beta3-is-out-max-sc-and-pd/1041/20

how to reproduce:

1. load the bufmfcc helpfile
2. change @startcoeff 1
3. run = all is good
4. run again = last coeff is DC
5. resinstantiate = good the first time, DC the following time

It is also the case in Pd so I reckon it is a core/algo issue

p

weefuzzy:
fixes #59 

Problem was that after `reset` the DCT object has wrong size (gives a handy assertion death in debug(


tremblap:
oops I should have assigned to self and try in Test compile indeed, sorry!

PR #62, opened 2021-11-06
Remove SIMD detection magic from CMake

weefuzzy:
Caused more trouble than it solved, especially for mac UB2. 

Passing `FLUID_ARCH` from the command line should still work. This just gets rid of the hidden magic that would try and set `march` and `mtune`. If we start using CMake presets, then that would offer a less hideous way to share some useful options in a less brittle way, maybe.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -181,7 +181,6 @@ if(APPLE)
 endif()
 
 #Apply any vector instruction flags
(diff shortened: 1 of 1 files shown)


jamesb93:
This looks good to me and I was able to build something that worked. Having some cmake presets would be grand too!

PR #65, opened 2021-11-16
Add local + core commit SHAs to version string

weefuzzy:
This uses `git log` to retrieve abbreviated SHAs for flucoma-core and the project being built, and appends them to the version string. 

I've used the [semantic versioning guidelines for including build meta data](https://semver.org/#spec-item-10)
>Build metadata MAY be denoted by appending a plus sign and a series of dot separated identifiers immediately following the patch or pre-release version. Identifiers MUST comprise only ASCII alphanumerics and hyphens [0-9A-Za-z-]. Identifiers MUST NOT be empty. Build metadata MUST be ignored when determining version precedence. Thus two versions that differ only in the build metadata, have the same precedence. Examples: 1.0.0-alpha+001, 1.0.0+20130313144700, 1.0.0-beta+exp.sha.5114f85, 1.0.0+21AF26D3—-117B344092BD.

### DIFF
--- a/script/flucoma_version.cmake
+++ b/script/flucoma_version.cmake
@@ -10,6 +10,19 @@ include_guard()
 
 find_package(Git REQUIRED)
 
+macro(getsha workingDir varName)
+  execute_process( 
+    COMMAND ${GIT_EXECUTABLE} log -1 --format=%h
+    WORKING_DIRECTORY ${${workingDir}}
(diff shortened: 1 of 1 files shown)


jamesb93:
This is gorgeous work:

image

Thanks for sorting this business out.

jamesb93 (approved):

PR #67, opened 2021-11-24
Toggle BLAS settings for compiling on APPLE

jamesb93:
This disables BLAS for eigen in the builds for APPLE.

In the future it should be revisited, especially with 3.4 Eigen which might solve the problem.﻿

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -170,7 +170,7 @@ endif()
 target_compile_definitions(FLUID_DECOMPOSITION INTERFACE EIGEN_MPL2_ONLY=1)
 
 if(APPLE)
-  target_compile_definitions(FLUID_DECOMPOSITION INTERFACE EIGEN_USE_BLAS=1)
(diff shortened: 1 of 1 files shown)


CMakeLists.txt
```diff
@@ -170,7 +170,7 @@ endif()
 target_compile_definitions(FLUID_DECOMPOSITION INTERFACE EIGEN_MPL2_ONLY=1)
 
 if(APPLE)
-  target_compile_definitions(FLUID_DECOMPOSITION INTERFACE EIGEN_USE_BLAS=1)
+  # target_compile_definitions(FLUID_DECOMPOSITION INTERFACE EIGEN_USE_BLAS=0)
```
weefuzzy, line 173: I think just delete this line. Individuals can always restore using BLAS from the command line (i.e. by passing `-DEIGEN_USE_BLAS=1`).

weefuzzy (changes requested):

PR #68, opened 2021-11-26
add chroma feature to NoveltySlice

g-roma:
This PR adds chroma to NoveltySlice. The new enum is "Spectrum", "MFCC", "Chroma", "Pitch", "Loudness", so code using pitch or loudness would be affected. Tested in SC. The SC class does not need modification, but the helpfiles on all platforms would need to be updated.
For using with tonal material larger kernel sizes than the default (3) are advised.

### DIFF
--- a/include/clients/rt/NoveltySliceClient.hpp
+++ b/include/clients/rt/NoveltySliceClient.hpp
@@ -16,6 +16,7 @@ under the European Union’s Horizon 2020 research and innovation programme
 #include "../common/ParameterConstraints.hpp"
 #include "../common/ParameterSet.hpp"
 #include "../common/ParameterTypes.hpp"
+#include "../../algorithms/public/ChromaFilterBank.hpp"
 #include "../../algorithms/public/DCT.hpp"
 #include "../../algorithms/public/Loudness.hpp"
 #include "../../algorithms/public/MelBands.hpp"
@@ -43,7 +44,8 @@ enum NoveltyParamIndex {
 };
 
 constexpr auto NoveltySliceParams = defineParameters(
-    EnumParam("feature", "Feature", 0, "Spectrum", "MFCC", "Pitch", "Loudness"),
+    EnumParam("feature", "Feature", 0, "Spectrum", "MFCC", "Chroma", "Pitch",
+              "Loudness"),
     LongParam("kernelSize", "KernelSize", 3, Min(3), Odd(),
               UpperLimit<kMaxKernelSize>()),
     FloatParam("threshold", "Threshold", 0.5, Min(0)),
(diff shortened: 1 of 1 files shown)


jamesb93:
Nice! Would it be possible to maintain as much of the enum as possible so that chroma becomes `feature 4`?

jamesb93:
I also just moved the PR destination to dev as this is where we'll be (hopefully soon) triggering the CI of nightlies.

g-roma:
Nice, thanks. Yes it could be moved, although it makes more sense like this to me. I would expect multivariate features like MFCC and Chroma to be easier to work with.

jamesb93:
I agree it makes more sense in the long run to have the multivariate descriptors together. I'm for reshuffling the enum and breaking patches in that case as its relatively easy to fix as long as we make users aware.

jamesb93 (approved):

weefuzzy (approved):
This is a breaking change for existing CCE code that has been using either Pitch or Loudness as an underlying feature. However, I think sucking that up (but making sure we announce it with big red letters) is better in the longer term than living with an Enum ordering that 'leaks' the order in which things were implemented. (IOW, I agree with Gerard, after some consideration)

This feels like about as much extra weight as can be added to this Client, IMO: the code is unattractively branch-y and magic number-y now, so I'd like to see two things in the future
1. A refactor of this Client to do the runtime dispatch better than if-else trees and switch cascades. 
2. The planned work on factoring out the novelty function into a lower level object (along with something to do the change point detection) so that folk can use this with their own combinations of features

jamesb93:
> 2. The planned work on factoring out the novelty function into a lower level object (along with something to do the change point detection) so that folk can use this with their own combinations of features

This should go elsewhere but while we're at the shops, would it look like:

feature buffer -> novelty

So you can calculate the novelty of any nd features?

weefuzzy:
> This should go elsewhere 
yes 😉 

> but while we're at the shops, would it look like:
> 
> feature buffer -> novelty

That's what I imagine

jamesb93:
😮

tremblap:
I'm ok to make it better and more consistent. but that implies 2 contradicting orders:

- the one proposed (which could be ok but it is still arbitrary)
- another one, in order of complexity for the users: energy is simpler. then fft, then we should put mel I reckon, then MFCC, then pitch then chroma
 
In all cases, I would not change the default kernel size though. We can leave a note in the help.

So this interface question is complicated. If we change the order now, we need to be prepared to change it again when we decide we want to add mel for instance, or any other descriptor. And then what is the rationale of multivariate together is ok but why not at the end?

jamesb93:
> I'm ok to make it better and more consistent. but that implies 2 contradicting orders:
> 
> * the one proposed (which could be ok but it is still arbitrary)
> * another one, in order of complexity for the users: energy is simpler. then fft, then we should put mel I reckon, then MFCC, then pitch then chroma
> 
> In all cases, I would not change the default kernel size though. We can leave a note in the help.
> 
> So this interface question is complicated. If we change the order now, we need to be prepared to change it again when we decide we want to add mel for instance, or any other descriptor. And then what is the rationale of multivariate together is ok but why not at the end?

I think keeping it as close to its former self is better than breaking it entirely. Rule of least surprise or something like that. I don't find "energy" (loudness I assume?) to be simpler _in the result it produces_ and in fact I often find it work not that well. I get the most tuneable results with MFCC and Spectrum, so for me having multivariate features that allude to the Foote paper should be preferentially located as the first enums. In that sense the order becomes less arbitrary and more of a conceptual pathway as one might explore the different features. This is of course all moot in the face of breaking the interface apart into many objects anyway potentially.

tremblap (approved):
After a quick discussion with @weefuzzy he made me saw the light. Let's get this approved and go forward and ignore my melband addition

tedmoore (approved):

jamesb93:
With 4 🆗 I merged :) 🚀

PR #71, opened 2022-01-06
added color schemes and credits for them

tedmoore:
I implemented one of these color schemes (https://colorcet.com/) as an option for the spectrogram of SuperCollider's FluidWaveform (other color schemes are of course possible, there are many in this folder). Of course this needs to be cited, so I put that in the folder as well.

Thoughts about using these files?
About storing them in the Audio Files folder?
About the citation?

Screen Shot 2022-01-06 at 13 57 28


jamesb93:
I think the attribution is fine and you follow exactly what they ask you to do. There are a few things though:

1. If we're going to use these color schemes there needs to be a path forward for creating that feature in Max's `fluid.waveform`.
2. If we do agree on implementing colour schemes then storing that information in a way that makes sense either for all of them (such as storing them in core). I don't see storing them in CSV's as the way forward for Max, it would be easier and less files to manage to encode them as a `Map()` in the javascript code directly which... nullifies the point of making a cross platform place to keep them.

I would just opt for implementing them on a CCE-basis as the implementation will change quite drastically between them. If no-one agrees with me then I would at least opt for not storing these in the AudioFiles folder but something called "resources".

tremblap:
1. I like this idea very much - sexy looking and options, what is there not to like :)

2. I agree that csv in audiofiles is messy

So maybe we could:
- put them in core as dependencies and retrieve them at build time wherever they need to be (pros: no duplicate, cons: code heavy for the wrappers)
or
- put them in each CCE in the right place (pros: more idiomatic, cons: duplicate)

I think the credits should go inside the doc at that attribute with a link.

jamesb93:
> * put them in core as dependencies and retrieve them at build time wherever they need to be (pros: no duplicate, cons: code heavy for the wrappers)

I just want to reiterate that CSV is portable, but not the ideal format for Max which already nullifies 33% of the benefits of having the code in one place. Given that research and colour palette has been around and unchanged for a long time putting it in each place in a format that makes sense natively makes sense and reduces code management.

weefuzzy:
I should have read your other PR properly. The missing CSV makes sense now. 

I think I agree with @jamesb93 here that there's no clear benefit to having these in core (tbh, now that we're so trigger happy with repos, I'd be more in favour of having the audio files *out* of core and in their own home), and that for JS plotting, I'd be inclined to JSON-ify the CSVs instead. 

FWIW, I don't think it will be a massive amount of work to bring the feature to `fluid.waveform~`, whenever one of us has the time or inclination (basically `Layer.drawImage()` in `display-jsui` needs access to some mapping function that we can pass in from above.  

(I think this is broadly in line with my position that I don't especially mind if the Max and SC graphical widgets differ in interface and functionality, because I see them as ancillary support to the main event, and so think that local idiomacy is most important for them)

tremblap:
I am happy to keep this in each repo in a bespoke wat as it is small, imported and low maintenance, and needs different format.

I strongly disagree to remove the Audio from core as it is shared and core, but that is another discussion for elsewhere.

tedmoore:
Cool. I will put them as CSVs in the SC repo in a folder called resources.

PR #75, opened 2022-01-19
Enhance/versioning

weefuzzy:
Here's a bash at changing the versioning as discussed 

1. The overall version bits are set by a human in flucoma-core/FlucomaVersion.cmake 
2. A file called `flucoma.version.rc` is generated in core's root that has the non-sha extended version string that can be used to make a tag

The generated C++ file is now a translation unit (.cpp) rather than a header. This means we can compile it separately as a library and then link with the client at the end – crucially this means that a change to the version string doesn't entail any binary change to the client itself (as well as speeding up rebuilds where nothing has changed).


jamesb93:
excellent - mind if we meet briefly on this to discuss how it can integrate with CI builds?

weefuzzy:
Sure, I'll ping you in 5?

jamesb93:
okidoki!

jamesb93 (approved):

jamesb93:
Approve wholeheartedly - this is going to simplify a lot.

tremblap:
go for it tiger. I was not asked officially but here goes my vote for it :)

jamesb93:
I've tested grabbing the tag from the .rc file which is only a `cat` away, as well as whether or not the tag bakes into the exec which it does so I'm more than happy :)

PR #23, opened 2021-08-13
EnvelopeGate: Address performance by avoiding (so much) copying

weefuzzy:
resolves #22 

Checked in SC and Max. SC tests pass. 

In the interests of simplicity in already complex code, the current implementation updates its input and output buffers by block copying the old buffer content to its new positions. This introduces a great deal of overhead, which grows markedly as `minLengthAbove` / `minLengthBelow` get bigger, and pegs out at 100% CPU for reasonable-seeming values. 

Rather than copying the data around, here I instead do the conventional ring buffer thing of moving read / write heads through the buffers. The downside, obviously, is that operations that were previously on contiguous ranges now sometimes have to be split into two operations, where the range crosses the edges of the buffer. This adds complexity / decreases clarity. On the upside, it means that performance no longer degrades steeply when using these attributes. 

I've done everything inline here in the name of applying the minimal changes needed to fix the problem and pass tests. However, happy to go off and consider if there's a way of hiding some of this complexity away so that the intention of the algorithm code can remain clear(er). The three main horrible bits are `refineStart` (where the search for the next minimum may have to be spread over two separate reads of the input buffer), and the two blocks in `processSample` that set ranges of the outputBuffer.

PR #82, opened 2022-01-20
fix issue #23 - zeroing the output vector in the mTracker loop

tremblap:
This test code should show the difference.

```
x = {|nband=10| FluidChroma.kr(WhiteNoise.ar(0.2), nband, normalize: 1, maxNumChroma: 12).poll(1)}.play
//10 wiggles

x.set(\nband, 4)
//sample and hold instead of 0 for bands 5-10 (now fixed)

// ====

x = {|nband=10| FluidMelBands.kr(WhiteNoise.ar(0.2),nband,maxNumBands: 10).poll(1)}.play
//10 wiggles

x.set(\nband, 4)
//sample and hold instead of 0 for bands 5-10


// ====

x = {|nband=10| FluidMFCC.kr(WhiteNoise.ar(0.2),nband,maxNumCoeffs: 13).poll(1)}.play
//10 wiggles

x.set(\nband, 4)
//sample and hold instead of 0 for bands 5-10

s.reboot
```

### DIFF
--- a/include/clients/rt/ChromaClient.hpp
+++ b/include/clients/rt/ChromaClient.hpp
@@ -90,6 +90,7 @@ public:
     {
       mMagnitude.resize(get<kFFT>().frameSize());
       mChroma.resize(get<kNChroma>());
+      output[0].fill(0);//zero the output
       mAlgorithm.init(get<kNChroma>(), get<kFFT>().frameSize(), get<kRef>(),
                       sampleRate());
     }
--- a/include/clients/rt/ChromaClient.hpp
+++ b/include/clients/rt/ChromaClient.hpp
@@ -90,7 +90,6 @@ public:
     {
       mMagnitude.resize(get<kFFT>().frameSize());
       mChroma.resize(get<kNChroma>());
-      output[0].fill(0);//zero the output
       mAlgorithm.init(get<kNChroma>(), get<kFFT>().frameSize(), get<kRef>(),
                       sampleRate());
     }
@@ -103,6 +102,7 @@ public:
         });
 
     output[0](Slice(0,get<kNChroma>())) = mChroma; 
+    output[0](Slice(get<kNChroma>(), get<kMaxNChroma>() - get<kNChroma>())).fill(0); 
   }
 
   index latency() { return get<kFFT>().winSize(); }
--- a/include/clients/rt/MFCCClient.hpp
+++ b/include/clients/rt/MFCCClient.hpp
@@ -103,7 +103,6 @@ public:
       mMagnitude.resize(get<kFFT>().frameSize());
       mBands.resize(get<kNBands>());
       mCoefficients.resize(get<kNCoefs>() + !has0);
-      output[0].fill(0);//zero the output
       mMelBands.init(get<kMinFreq>(), get<kMaxFreq>(), get<kNBands>(),
                      get<kFFT>().frameSize(), sampleRate(),
                      get<kFFT>().winSize());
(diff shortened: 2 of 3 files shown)


weefuzzy (changes requested):
This isn't quite what I suggested. Filling `output[0]` only when the parameter value changes makes the assumption that whatever `output[0]` points to is the same thing between invocations of `process()`. I don't think that's a guarantee we make. 

Instead, my suggestion was that you fill the remainder of the output vector when the function returns by writing something like `output[0](Slice(get<kNBands>()).fill(0)` after the chroma values have been assigned.

weefuzzy:
The problem is here: 

https://github.com/flucoma/flucoma-docs/blob/0e970cb811e59e541ec41348e69c9302af254a45/flucoma/doc/raw/driver.py#L15-L18

maybe try resolving the path or somethng?

tremblap:
I am not yet good enough to understand why process() would not get the same output[0] but I have an intuition... but adding 0s every time seems heavy-handed, no? I'll give it a spin.

tremblap:
your solution works, except when kNXXX == kMaxXXX, which spits all 0s

before I do a if, or put the zeroing in front, is there anything I could do better in the world of FluidTensor(View)

weefuzzy:
`output[0]` is, at heart, a pointer and the Client knows nothing about the provenance of that pointer. It could be something provided by the host *wrapper* or the host itself. In any case, assuming that it remains pointing to the same thing between invocations involves also making assumptions about how hosts and wrappers behave, and how the lifetime of the client might relate to, e.g., the lifetime of a particular job in the host. It will happen to work in real-time usage in Max, PD, SC at the moment. Pretty sure it will break in the `Buf` versions. 

Writing a few zeros isn't at all heavy handed relative to the amount of other work that's going on.

tremblap:
thanks again for the thorough explanation as usual. I'm still puzzled why the Slice of length 0 is zeroing all but i can work around

weefuzzy:
> your solution works, except when kNXXX == kMaxXXX, which spits all 0s
> 
> before I do a if, or put the zeroing in front, is there anything I could do better in the world of FluidTensor(View)

Hmmph. Maybe pass the slice size explicitly 

output[0](Slice(get<kNBands>(),get<kMaxBands>() - get<kNBands>())).fill(0)

tremblap:
This works:
    output[0](Slice(get<kNChroma>(), get<kMaxNChroma>() - get<kNChroma>())).fill(0);

and this works too and I find it clearer conceptually:
    if (get<kNChroma>() != get<kMaxNChroma>()) output[0](Slice(get<kNChroma>())).fill(0);

maybe it should be < instead of !=

weefuzzy:
I'd rather not add extra branches into a code hot path where we don't need them: the point is that this condition is checked anyway, so no need to do it twice.

tremblap:
You are hard on me. So it is better to send a 0size Slice (when the subtraction is 0)  than a fork? I'll follow your lead and learn strange conventions :)

weefuzzy:
Yes, otherwise the branch predictor's job is just more complex for no benefit. The thing iterating the slice is going to be doing something like `while(current != end){...}` anyway, so a replicating this branch is redundant.

tremblap:
I was just trying to remove a subtraction :)

weefuzzy:
Branching is worse ;-)

tremblap:
Properly working now

weefuzzy (approved):

PR #83, opened 2022-01-20
Tests for `include/data`

weefuzzy:
This is the first tranche of tests, starting with the container-like classes in `include/data` (`FluidTensor`, `FluidTensorView`, `FluidTensorSupport`, `FluidDataSet`). Coverage is not yet complete...

* Everything is runnable through CTest, via CMake so these should be CI friendly 
* Introduces new dependencies: [catch2](https://github.com/catchorg/Catch2)  and [ApprovalTests.cpp](https://github.com/approvals/ApprovalTests.cpp) (only when testing is enabled) 
* Some additional plumbing done to support testing for `assert` deaths and compile time stuff. Both of these work, but could be better 

I think this covers the spectrum of types of testing we'll need. Catch is pretty easy to use for regular tests (and does a bunch of other stuff I haven't really investigated properly yet), and the approval testing will be invaluable for those things that are just hard to test in other ways (though it takes some getting used to). 

To enable tests pass `-DFLUCOMA_TESTS=On` to CMake. This will lengthen the configure time quite markedly, so it's not on by default. Tests should then be build when you run `make` or whatever (again, lengthening the build time appreciably). 

To run, you can call `ctest .` or `make test` or `ninja test`  rom your build folder. Depending on your generator it might want more flags passing (e.g. in Xcode). 

* The compilation tests take much longer than the other ones, because they're trying to build there and then. I think there's probably some exclusion flag we can use with CTest when we don't want to be bothered with them
* Compilation tests should be sparingly used: they're really only needed for testing `static_assert`s: there are facilities within Catch that can help with other types of compile time tests. 
* Getting the most out of the approval tests requires having a compatible diff & merge tool installed (although there are ways to override that, and it's automatically suppressed in CI contexts). Weirdly Apple FileMerge isn't such a thing, so I've ended up installed kdiff3. This only applies when you're making a new approval test, or breaking an existing one. The rest of the time they should just work.


tremblap:
Congrats! Let me know if/how I can help. IIUC it changes nothing for -DFLUCOMA_TESTS=OFF or is there anything to check out for?

weefuzzy:
Indeed, not passing `-DFLUCOMA_TESTS=ON` or setting to something falseish should leave other things unchanged.

tremblap:
Let me know anything you would need me to check.

weefuzzy:
Checking out and confirming that you can work as normal without tests enabled would be useful, but not as a high priority. There's nothing in the tests to intersect hugely with your day to day at the moment, but at some point I'm sure you'll need to know how to run them. 

I guess the next job here is to get a workflow set up to run them so we can get the CI train a-chugging.

tremblap:
It all compiles fine, no more warning than usual. I take that it is off by default since I literally pulled that branch in the tip of dev and it all went well.

weefuzzy:
Great, cheers. Will merge then.

PR #87, opened 2022-01-28
Amend upper frequency limit of mel bands in NoveltySlice to 20kHz

### LINKED ISSUE #86: NoveltySlice w/ MFCCs: Mel frequency range correct?
@g-roma  

In `NoveltySliceClient`, `mMelBands` is initialized with a frequency range of 20Hz to 2kHz. Is this right, or did you mean 20kHz? 

https://github.com/flucoma/flucoma-core/blob/1e3148e163474c214df6a34cc2063d25cf91b406/include/clients/rt/NoveltySliceClient.hpp#L111-L112

weefuzzy:
closes #86 

This is a trivial change, fixing what seems to be a typo in the `NoveltySliceClient`. However, existing example code and stuff you might all be working on that uses novelty slicing with MFCCs will possibly be affected and need re-tuning. Hence a PR rather than just changing it. 

The expected results on the test will certainly need to change.

### DIFF
--- a/include/clients/rt/NoveltySliceClient.hpp
+++ b/include/clients/rt/NoveltySliceClient.hpp
@@ -111,7 +111,7 @@ public:
     else if (feature == 1)
     {
       mBands.resize(40);
-      mMelBands.init(20, 2000, 40, get<kFFT>().frameSize(), sampleRate(),
+      mMelBands.init(20, 20e3, 40, get<kFFT>().frameSize(), sampleRate(),
                      get<kFFT>().winSize());
       mDCT.init(40, 13);
       nDims = 13;


tedmoore (approved):

tedmoore:
thanks for the heads up!

tremblap:
happy to change, although the 5k idea of @g-roma seems to make sense too. we could also change the number of mfccs to 20 instead of 13 if we go higher def?

jamesb93 (approved):
I prefer 13 MFCCs as the default so think this is ready to merge.

tremblap:
The default is also 5k if you read @g-roma right... if we're breaking this, we could try a few options on ranges. Do we really care about 10k to 20k for instance, or sharp changes. 20 bands is also common for music material, etc

weefuzzy:
I think it ought to have the same behaviour as our `MFCC` object does by default, which is 20k. I can imagine 5k being used for speech research where sampling at 11k was SOP for quite some time, and perhaps for things where you can be sure that high frequency energy isn't that structured, but I don't think it's a useful restriction for a change detector where the control is fixed (and will stay fixed, before anyone gets any ideas). 

I don't think there's anything to be gained by embarking on a adventure of experiments correcting something that was a mistake to begin with. When we have a separate object for making novelty features, people can freestyle as they please.

tremblap:
ok let's park optimisation indeed

tremblap (approved):

tremblap:
@weefuzzy this would be ready to merge in right? When you do so, feel free to make an issue of the UT not working and putting my face on it

PR #90, opened 2022-02-08
Tests for framing and overlap add bits

weefuzzy:
This adds tests for the input framing and overlap-add in `clients/common` 

If one of you could verify that it works for you and then merge, that'd be ace (because I've got another one to come straight away!)

To test that it works configure and build with `FLUCOMA_TESTS` enabled, and from the build folder run `ctest .`

### DIFF
--- a/tests/CMakeLists.txt
+++ b/tests/CMakeLists.txt
@@ -127,6 +127,9 @@ target_link_libraries(TestFluidTensorDeath PRIVATE
 add_test_executable(TestFluidTensorView data/TestFluidTensorView.cpp)
 add_test_executable(TestFluidTensorSupport data/TestFluidTensorSupport.cpp)
 add_test_executable(TestFluidDataSet data/TestFluidDataSet.cpp)
+add_test_executable(TestFluidSource clients/common/TestFluidSource.cpp)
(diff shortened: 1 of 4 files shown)


tremblap:
Sorry for my ignorance - I pull the PR and compile the core only or in a cce? I've never done core alone but if that test flag works on all of them then I can compile SC with it

weefuzzy:
> Sorry for my ignorance - I pull the PR and compile the core only or in a cce? 

Core only was what I was imagining. However, it might be redundant because the CI has actually already done exactly what  I suggested (the windows blockage is cleared on dev now, but we can see above that the new tests built and passed on mac and linux).

> I've never done core alone but if that test flag works on all of them then I can compile SC with it

Might work! 🤷

tremblap:
so I did the usual with -DFLUCOMA_TESTS=ON - it pulled the various test frameworks and compiled them. Then from the build folder I tried `ctest .` but it found no tests. nor from the build of my core repo. so I'll try compiling the latter.

weefuzzy:
From the looks of it you'd need to descend into `build/_deps/flucoma-core-build` and do it from there

tremblap (approved):

tremblap:
and I should add - this is much sexier than the SC version

PR #91, opened 2022-02-08
Test/sc ports/slicers

weefuzzy:
Ports of the SC unit tests for slicers. These bypass the clients and work on the algorithms directly, so there's less stuff in the way. 

Includes useful plumbing for loading audio data for tests


tremblap:
I'm not sure what I have to do here except the new novelty mfcc values for the SC and C++ UTs, right?

weefuzzy:
Right. That shouldn't affect this PR. 

I need to find out why some tests are timing out on Windows, but that's unconnected :shrug:

weefuzzy:
Faulty tests fixed. I'd be inclined to merge this in now, seeing as the CI has run the tests themselves and now pronounces that all is well

tremblap (approved):
it worked for me locally - I haven't check performance of code but that is not part of the quesiton iiuc so go for it if you are happy with the windows fix

PR #93, opened 2022-02-09
bump Eigen version

weefuzzy:
Bump Eigen version to 3.4

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -52,10 +52,11 @@ FetchContent_Declare(
 
 FetchContent_Declare(
   Eigen
+  GIT_SHALLOW TRUE
   GIT_REPOSITORY https://gitlab.com/libeigen/eigen
   GIT_PROGRESS TRUE
-  GIT_BRANCH "3.3"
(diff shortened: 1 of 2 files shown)


tremblap:
this is a good moment to test the manual nightly of SC. I'l ltry to find the trigger button

weefuzzy:
you'll be searching a while because I haven't put it on yet! Also, this is on a PR branch: it wouldn't feature in the nightly until being merged...

tremblap:
ok so let's merge then - I cannot test locally without having to download it all which will take ages - if we merge to core we'll see the tests you run there and see if that works, then it'll be fun to see in action in the next nightlies, on which I'll run the UTs and expect a sign error (and will try on all 3 oses - then I'll change the UT to check for absolute value?)

weefuzzy:
The tests have already run on the PR (where it says All Checks Passed).

tremblap:
just got the nightly I was hoping for, will try to run while I do other things

tremblap:
ok as confirmed, the phase of PCA is now inverted. i'll try on Windows and Linux before modding the ut

PR #96, opened 2022-02-11
Fix/slicer blocksize dependency 

### LINKED ISSUE #95: `OnsetSliceClient`: results can depend on host blocksize
This probably goes for `NoveltySliceClient` too. 

## Problem 
The results of slicers are meant not to depend on the host blocksize. The code in `OnsetSliceClient` (resp. NovelitySlice) that deals with positioning spikes into output frames doesn't handle the case where the hop size is smaller than the block size, and also doesn't divide neatly into the block size. IOW words, assuming a power-of-two block size, a hop of 32 would give consistent results, whereas a hop of 40 wouldn't. 

## Workaround 

Um, don't use weirdass hop sizes. 

A fix PR is hot on the heels of this issue. The solution is to make the `frameOffset` part of the client state so that it can correctly track across block boundaries that don't tessellate with analysis frame boundaries. However, it might be better if the affected clients were able to pull this information directly from their `BufferedProcess` objects, which already implement this behaviour correctly. 

## Reproduction 
This supercollider code reproduces the problem 
```js
// Runs the same material through FluidOnsetSlice at different server block sizes
// the results should be the same each time 
(
var condition = Condition(); 
var results = Dictionary();
var synthFunc = {
    var sig = PlayBuf.ar(1,b,loop:1);
    FluidOnsetSlice.ar(sig,metric:8,threshold:0.3,hopSize:50)    
}; 

var recvFunc = {|a|    
    results.put(s.options.blockSize,a) ; 
    condition.unhang; 
    ("Done" + s.options.blockSize).postln;      
}; 

fork{    
    s.options.blockSize = 64;
    s.waitForBoot{
        b = Buffer.read(s,FluidFilesPath("Nicol-LoopE-M.wav"));
        s.sync; 
        synthFunc.loadToFloatArray(1,s,action:recvFunc); 
    }; 
    condition.hang; 
    s.quit; 
    s.options.blockSize = 256; 
    s.waitForBoot{
        b = Buffer.read(s,FluidFilesPath("Nicol-LoopE-M.wav"));
        s.sync; 
        synthFunc.loadToFloatArray(1,s,action:recvFunc); 
    }; 
    condition.hang; 
    s.quit;
    results[64].indicesOfEqual(1.0).postln; 
    results[256].indicesOfEqual(1.0).postln; 
}
)
```

weefuzzy:
fixes #95 

Using the test in the issue can confirm that now with both `OnsetSliceClient` and `NoveltySliceClient` you get the same results for a given hop size independent of the host block size. 

The existing SC tests aren't sensitive to this issue, because of how the NRT wrapper is currently implemented (it was changing this that made me discover the problem). However, both sets still pass here with this fix applied (so I'm less likely to have broken anything else).

### DIFF
--- a/include/clients/rt/NoveltySliceClient.hpp
+++ b/include/clients/rt/NoveltySliceClient.hpp
@@ -153,7 +153,7 @@ public:
     RealMatrix in(1, hostVecSize);
     in.row(0) = input[0];
     RealMatrix out(1, hostVecSize);
-    index      frameOffset = 0; // in case kHopSize < hostVecSize
+    
     mBufferedProcess.push(RealMatrixView(in));
     mBufferedProcess.process(
         windowSize, windowSize, get<kFFT>().hopSize(), c,
@@ -184,11 +184,15 @@ public:
             mLoudness.processFrame(in.row(0), mFeature, true, true);
             break;
           }
-          if (frameOffset < out.row(0).size())
-            out.row(0)(frameOffset) = mNovelty.processFrame(
+          if (mFrameOffset < out.row(0).size())
+            out.row(0)(mFrameOffset) = mNovelty.processFrame(
                 mFeature, get<kThreshold>(), get<kDebounce>());
-          frameOffset += get<kFFT>().hopSize();
+          mFrameOffset += get<kFFT>().hopSize();
         });
+
+    mFrameOffset =
+        mFrameOffset < hostVecSize ? mFrameOffset : mFrameOffset - hostVecSize;
+
     output[0] = out.row(0);
   }
 
@@ -203,6 +207,7 @@ public:
   void reset()
   {
     mBufferedProcess.reset();
+    mFrameOffset = 0; 
     initAlgorithms(get<kFeature>(), get<kFFT>().winSize());
   }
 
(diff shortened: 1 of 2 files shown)


tremblap (approved):
This works on all the use cases I have, as well as the example code of bug #95.

tremblap:
I tested with both flags on the compiler (optimised to local machine as well as CI settings). I also want to rave about the test code. I am learning, yet again.

weefuzzy:
Excellent. We should perhaps add some variation of the SC test code in #95 to the SC tests to guard against future regression (or make a C++ equivalent).

tremblap:
Let's merge this guy, and add a note. I am not certain where it makes more sense but if in SC then we'll need to play with the innards of Cameli's UTs code to enable comparison of block sizes. What could be fun is that all algo should be run at 2 blocksizes ... we could remove all tests that are done under the hood and keep only client/cce stuff in the client SC- UT to streamline the process

weefuzzy:
> we could remove all tests that are done under the hood and keep only client/cce stuff in the client SC- UT to streamline the process

That's certainly something to work towards, but for the moment having the tests duplicated is at least a useful way of indicating if something has gone screwy in client vs algorithm. *In pricniple* we would want the SC tests only to have to concern themselves with verifying our sclang code and communication with the server, but that's a journey...

Issue #94, opened 2022-02-09
make PCA polarity predictible


tremblap:
We could help our SVD to be predictible in polarity like scikit learn does.

https://github.com/scikit-learn/scikit-learn/blob/39c341ad91b545c895ede9c6240a04659b82defb/sklearn/utils/extmath.py#L710

PR #98, opened 2022-02-14
add pca whitening

g-roma:
This PR adds whitening to PCA. It changes the storage format of FluidPCA, sotring the "explained variance" instead of the singular values. I will send a PR to the flucoma-sc repository in short for the SC classes. I hope I have not missed anything from recent updates, but it is possible, as this has been in the pipeline for some time.


include/algorithms/public/PCA.hpp
```diff
@@ -35,47 +36,66 @@ class PCA
     MatrixXd         X = (input.rowwise() - mMean.transpose());
     BDCSVD<MatrixXd> svd(X.matrix(), ComputeThinV | ComputeThinU);
     mBases = svd.matrixV();
-    mValues = svd.singularValues();
+    mExplainedVariance = svd.singularValues().array().square() / (X.rows() - 1);
```
weefuzzy, line 39: I don't follow why we need to now store the explained variance rather than the singular values. Looks like it could equally well be calculated on demand by `getExplainedVariance`?

tremblap:
I'll leave you two to decide what to store but ideally we store something that allows for backward compatibility if possible and flexibility to whiten or now, right?

include/algorithms/public/PCA.hpp
g-roma: I guess the idea is to move the calculation to the training instead of every time you process or, worse, procesFrame, so in all it made more sense to me

weefuzzy: I'm uneasy about making a breaking change on that basis, especially if we don't actually know if there's a performance problem there yet (presumably the square root which is still in `process` and `processFrame` could dominate the cost). If there is a performance problem caused by doing it on demand, I'd be more inclined to cache the result during training rather than invalidate any saved PCA models that folk have already made.

g-roma: Note that the square root here is done on an array of size k, where as the operations for getting the explained variance work on the full dimensionality. The idea of keeping explainedVariance follows PCA in sklearn. There both the singular values and explained variance are kept. Here, I was not sure I would want to keep the singular values unless we have a use for them. Clearly, it is cheaper to keep and use the explained variance.
I did not weigh breaking current code against the added complexity as this is still in beta, but I am not too fussy about it, we can calculate the explained variance on the fly (or on loading the JSON in addition to training) if the priority is to keep the JSON format.

tremblap:
here we seem to have a breaking change, and I think we might as well implement a fix for #94 at the same time so we break PCA's back compatibility once of our users. Does it make sense?

weefuzzy:
We've established that the breakage here is avoidable, so I'd be inclined to avoid it. Nothing has been done about #94 yet, and I have  other irons in the fire, so no reason to let that hold this up.

tremblap:
ah I misread - then indeed let's decouple the 2 things.

g-roma:
Actually, computing the explained variance for whitening requires the size of the training data, so I don't think we can do it without breaking the JSON format.

weefuzzy:
> Actually, computing the explained variance for whitening requires the size of the training data, so I don't think we can do it without breaking the JSON format.

I'd still be inclined to minimize breakage by adding an new key for `N` to the JSON representation that would default to something if missing on read. That way old models will still work for dimension reduction, even though they'd be wrong for variance estimates, but there would be a trivial path for correcting this if the user still has a record of the size of the training set.

weefuzzy (approved):
Many thanks, this seems to be working fine.

weefuzzy:
@g-roma thanks also for absorbing the merge conflict from `dev` with the addition of inverseTransformPoint. 

Now that we have whitening I need to update this method to account for it (so whitened points are also invertible). Do you mind if I do it on this branch?

tremblap:
see agenda of meeting of 28/3 for everyone's todos r.e. child prs on max,pd,sc,doc

Issue #18, opened 2021-06-26
MLP::getParameters() is not correct


AlexHarker:
MLP::getParameters() is not correct

The weights and biases cannot return anything useful as they are not passed in by reference, so the method sets local versions of these only. Also, the extents need to match, so apparently (according to @weefuzzy this may need to be done with reset().

g-roma:
Not sure I see the issue here, the weights and biases are passed as FluidTensorView instances, this is not a container itself but manages a pointer to the actual storage. The assignment operator then copies the data to the FluidTensor allocated by the caller, which may be a bit confusing, but is a pattern used generally in the code base. The sizes are obtained from the MLP instance itself, this method is used in FluidJSON to store the MLP parameters as JSON.

AlexHarker:
I understand what you are saying, but this is currently not at all obvious from reading the code. Looking at asFluid() it is very hard to disentangle.

If you are confident that this is correct (and indeed the usage in FluidJSON.hpp and my own experience would seem to confirm that) then I'm happy for this to be closed.

As a side issue there's definitely something to think about should you wish the code to be navigable and useable by mere mortals - even if that's just a comment on the methods that use this return pattern, as it's very unlikely (IMHO) even an experienced coder would understand what is going on.

AlexHarker:
To. be more constructive - the simplest solution to the documentation issue (but one that likely @weefuzzy might rightly not be happy about) would be to rename types - if "Views" were "Ptrs" or something like that then this behaviour would become much less surprising from the viewpoint of the method, although the code within the method would still be non-obvious, as setting the value of a pointer type (the operation in the method is = ) generally doesn't result in changing the underlying data of the original pointer.

weefuzzy:
Here obviously isn't the place for thrashing through alternative designs, but I will not that this source of confusion is more my fault than @g-roma's. 

We all agree that having `operator=` do a deep copy is confusing (although this is also what `Eigen::Map` does, so at least we're consistently confusing). Our experience so far has been pretty clear that having some sort of operator to do a multi-dimensionally range-checked deep copy is tremendously handy, both for coding quickly and for catching all sorts of bounds errors early. Perhaps a different operator would be clearer, whilst retaining the convenience.  

[...]

PR #100, opened 2022-02-14
FluidTensorView deep copy gets distinct operator

weefuzzy:
Apropos a discussion in #18 some time ago, this replaces the confusing use of `operator=` in `FluidTensorView` and `FluidTensor` for deep copying from a `FluidTensorView` and instead uses `operator<<=`. @AlexHarker and @g-roma may have opinions, or not. 

* `operator<<=` becomes a shorthand for  `std::copy(src_view.begin(), src_view.end(), dest.begin())`, where `dest` might be either a `FluidTensor` or a `FluidTensorView`. It also asserts in debug builds that `source` and `dest` have the same extents. This check, IME, has been one of the primary benefits of having a deep copy operator, besides the syntactic sugar, as it's caught a great many potential bugs early. 
* `FluidTensor::operator=` remains, but only applies to copies / moves from `FluidTensor` and should (still) have the same semantics as container copy / move assignment in the STL (viz. reallocation will happen as needed) 
* `FluidTensorView::operator=` becomes a shallow copy / move, again bringing it into line with the STL (i.e. how `std::span` behaves).


AlexHarker:
That sounds good to me - it's easier to follow the code if things called views are not able to alter the underlying data with an = operator. What you've chosen will certainly make people think harder and it seems like a good operator to chose.

PR #106, opened 2022-02-24
Enhance/optional message args

weefuzzy:
See https://github.com/orgs/flucoma/teams/flucoma-dev/discussions/3


weefuzzy:
checks failed because it tried to run the tests on `tl::optional` without building them. will strop that happening

tremblap:
this is great. A few questions

- at first I was surprised that the attrui was not updating but I reckon this is the desired behaviour - the 2nd message is optional
- the radius part of the kdtree query brings an interesting question - should we add another (3rd) argument there?

this looks like a very promising development for all the artificial attributes, i.e. all those unliked to the state of the object

tedmoore:
Just reviewed the SC version (https://github.com/flucoma/flucoma-sc/pull/77).

@weefuzzy This is great. I think it makes a lot of sense.

PR #108, opened 2022-02-25
PoC: Enhance/choices param

weefuzzy:
@tremblap @jamesb93 @tedmoore @jdchart 
Adds a `choices` parameter type allowing for selections from fixed pool of options, e.g. in `BufStatsClient` or `SpectralShape`. Like an `enum` the client author supplies a list of strings. The combination of options selected by the user is encoded as a bit field, which makes it easy to check / count in the client. 

`BufStatsClient` is experimentally updated to add a new control, `select`, to try this idea out. I picked this as the more complex case (needing to account for derivatives in the output), so hopefully `SpectralShape` would be simpler. 

e.g if you have, say, `select = mean` and `numDerivs = 1` you'll get two outputs (mean and ∆-mean), whereas with `select = mean std` and `numDerivs = 1` you'd get 4, and so on. 

I think this makes the need for some sort of indexing scaffolding that @tedmoore suggested even more acute.  

Leaving as a draft as this is a proof of concept until we've played with it a bit, by which time it'll probably need rebasing. 

Accompanying PRs for docs, max, sc, pd and cli incoming.


weefuzzy:
https://github.com/flucoma/flucoma-max/pull/147
https://github.com/flucoma/flucoma-sc/pull/78
https://github.com/flucoma/flucoma-pd/pull/31
https://github.com/flucoma/flucoma-cli/pull/26
https://github.com/flucoma/flucoma-docs/pull/90

jamesb93:
In the abstract I really like this idea. It will reduce the need for curating statistics (often incredibly repetitively) and make code more readable and understandable. Also will be easier to programmatically change what statistic(s) you care about on the fly or in a more complex processing chain.

jamesb93:
In the concrete I like this idea even more :) We need to document explicitly how the ordering works and that this is a _mask_, rather than an ordered cherry picking type situation. I vote merge ASAP and document even sooner.

tremblap:
this is great - the naming will have to be thought carefully for consistency - `select` vs `choice` vs `which` vs `output`

I think that spectralshape is the next contender for that and should help think about the vocabulary/interface but otherwise big thumbs up for me on this.

p

jamesb93:
> this is great - the naming will have to be thought carefully for consistency - `select` vs `choice` vs `which` vs `output`
> 
> I think that spectralshape is the next contender for that and should help think about the vocabulary/interface but otherwise big thumbs up for me on this.
> 
> p

`select` has some kind of allusion to bufselect, so perhaps thats a stronger case than `choice`. Though in a bubble, `choice` rings truer to its meaning.

tedmoore:
for what it's worth, i like `select`

tremblap:
I love interface discussion. @tedmoore did you get the right stats in SC? I tried max and pd and get strange behaviour - I've registered those in the max pr

tedmoore:
> I love interface discussion. @tedmoore did you get the right stats in SC? I tried max and pd and get strange behaviour - I've registered those in the max pr

Ah! As this is still a draft I wasn't sure it was ready for testing. @weefuzzy, is it ready for testing?

tremblap:
this correction has fixed the problem indeed... but it seems to have broken all the other objects!
Screenshot 2022-03-23 at 12 07 24

weefuzzy:
that's not happening here (and I can't see how one would cause the other) – try re-running cmake, cleaning etc. What architecture is the machine you're currently working on?

include/clients/common/ParameterTypes.hpp
```diff
@@ -126,6 +127,37 @@ struct EnumT : ParamTypeBase
...
+  
+  template <index... N>
+  constexpr ChoicesT(const char* name, const char* displayName,
+                  const char (&... string)[N])
+      : ParamTypeBase(name, displayName), strings{string...}, fixedSize(1),
+        numOptions(sizeof...(N)), defaultValue((1 << numOptions) - 1)
+  {
+    static_assert(sizeof...(N) > 0, "Fluid Param: No choice strings supplied!");
+    static_assert(sizeof...(N) <= 16,
+                  "Fluid Param: : Maximum 16 things in an choice param");
```
jamesb93, line 143: just to make a note here this still says choice and not select

include/clients/nrt/BufStatsClient.hpp
```diff
@@ -43,6 +44,7 @@ constexpr auto BufStatsParams = defineParameters(
     LongParam("startChan", "Start Channel", 0, Min(0)),
     LongParam("numChans", "Number of Channels", -1),
     BufferParam("stats", "Stats Buffer"),
+    ChoicesParam("select","Selection of Statistics","mean","std","skew","kurtosis","low","mid","high"),
```
jamesb93, line 47: SelectParam?

tedmoore: Maybe the "Param" bit is implied and therefore added noise? I might also suggest that it's not a Parameter, it's a statistic?

jamesb93: I think that Param is there in the C++ because its a C++ thing, not a user-facing thing to keep it consistent with the other things which are defined as parameters/attributes.

tremblap: I have deleted the whole build folder. I'll redo that.

jamesb93:
> btw `select` is growing on me - the case @jamesb93 did on bufselect (similar job similar name) is good - I just wonder how it'll port to spectralshape - although the same reasoning applies...

there is a basis on Max for it too after `rg` through Max's "source". Objects that use it in some flavour:

`mc.resize~ @select` (to only include certain channels in the output).
`jit.gl.handle @select_mode`

tremblap:
just to be certain: max 147 core 108 noDocs right?

tremblap:
docs needs 90

tremblap:
I'm doing Pd in parallel - so much compilation time on an old 4 processor machine...

include/clients/nrt/BufStatsClient.hpp
weefuzzy: There's no need for renaming the C++ type, I would argue

weefuzzy: Everything in that layer is contingent anyway

tedmoore:
This is working great for me. 

The warnings for `is not a recognised option` is excellent. Is there a way to throw a warning if a user specifies one of the options twice? I notice that this does cause some funky numbers (below). It appropriately returns only the three stats (instead of high twice or something), but he "high" values should not be all zeros (they are not all zeros when just "high" is requested).

```
[ high, high, mid, std ]
[ 81.231330871582, 35.592910766602, 15.353200912476, 9.4192552566528, 12.277660369873, 8.4405298233032, 8.9031448364258, 6.961977481842, 6.0791192054749, 4.573944568634, 4.4644193649292, 3.9180850982666, 4.1074557304382 ]
[ -16.664413452148, 24.666746139526, 11.945391654968, 14.825135231018, -2.1219320297241, -0.89657753705978, -5.1428756713867, -2.052295923233, -3.6580066680908, 0.65312850475311, 2.2503006458282, 1.4344000816345, 2.2939376831055 ]
[ 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0 ]
```

weefuzzy:
@tedmoore The problem was in the SC class – evidently wasn't thinking when I did the code to form the bitfield. 

I think we can warn on duplicates (not that they harm things now) with something like 
```js
[\high, \high, \mid ,\std].asBag.countsDo{ |item, count, i|
    if(count > 1) { ("Option '" ++ item ++ "' is repeated").warn}
}
```

?

jamesb93:
The change is incredibly good for UX and makes it simpler to grok your code when sharing/going back to it. Will need to update the relevant docs but I approve.

PR #114, opened 2022-03-09
Enhance/slicer detection curves

weefuzzy:
Adds new client that produce the detection function curve for Novelty, Onset and Amp Slicers (which involved refactoring the respective algorithms to separate curve from detection)


weefuzzy:
we probably want to think about naming a bit

jamesb93:
In principle I don't see why _any_ of these objects shouldn't be merged immediately, except for the fact that there is no documentation for them. I think that should be appended to this PR and I can offer to do this. 

One concern I have is that it might be hard to understand the connection these curves have to the objects which do the segmenting, particularly as each one may have its own idiosyncratic way of digesting the curve and returning a point of change. All in all, I think that's a communication and documentation issue and I believe there are compelling examples to be made like this:

## Kernel 7
image

## Kernel 21
image

For showing how an actual novelty curve changes with the kernel.

Another thought is how do these fit into documentation. Do we see them as their own beast, or something that assists with the slicers. In the latter case, I believe we could simply introduce a tab for the appropriate slicer's help file that weaves these objects in and especially as they share parameters.

jamesb93:
Not to mention how insanely good these tools will be for teaching!

tedmoore (changes requested):
I think once these have documentation in the RST they should be merged!

jamesb93:
> I think once these have documentation in the RST they should be merged!

and help files for PD Max too

weefuzzy:
> ok this doesn't compile on my mac - 114 on core, 151 on max - the error is:

Should be fixed now

tremblap:
@jamesb93 nice help visualisation in novelty - although I'm surprised by the peak-picking results but that might be a bug in the novelty :)

So we need to find the names, make the helpfiles in all CCEs, the docs (for the CLI too!) and this is good to go. I am opened to suggestions on the merge order, including a temporary out of sync between cces

jamesb93:
> the problem I see with adding anything after the feature is that we load it with meaning - you know are careful we've been in trying to keep them as agnostic as possible. I'd like to keep our pattern (also we are not providing curves in realtime but streams)
> 
> transient -> transientslice novelty -> noveltyslice
> 
> ampgate -> we don't have anything but loudness @filter 0 is the nearest of an energy curve/stream people get
> 
> now let's brainstorm for the problematic 2:
> 
> ampslice -> ampdiff ? ampchange ? ampdetrend?
> 
> onsetslice -> spectralchange (which is what it is trying to do as a stream/curve with 10 different metrics) ? spectraldiff? spectralflux (although that has already a meaning in descriptor literature)

I really see these curves as inseparable from the slicers and making a new taxonomy for them confusing both to use, remember and associate. Not to mention without the word curve~ at the end (or something), the `transientscurve~` becomes... `transients~` 🙅‍♂️.

FWIW @tremblap I see your point of view on the agnostic naming but I don't think its worth the cognitive overhead for everyone involved as well as loading it in such a small way. That's my take on it.

If we were to go that route my vote would be (excluding transients which I cannot think of anything for):

`fluid.bufnovelty~`
`fluid.bufampdiff~`
`fluid.bufspectraldelta~`

tremblap:
> I really see these curves as inseparable from the slicers

that is exactly why I think we need to not use curve. if they are that linked it should be easy to users to make that link, and would avoid limiting them to that in more divergent users - they are time series after all.

continuing on the brainstorm: I love spectraldelta but is it too geek?

for ampdiff I thought of ampcrest - in relation to this https://en.wikipedia.org/wiki/Crest_factor - but again maybe that is too geek

let's keep thinking. interface questions are worth it.

jamesb93:
> continuing on the brainstorm: I love spectraldelta but is it too geek?

probably. I thought of it because of the `delta` attribute in the object. `diff` gets to the same goal in less characters too.

> that is exactly why I think we need to not use curve. if they are that linked it should be easy to users to make that link, and would avoid limiting them to that in more divergent users - they are time series after all.

you're flipping the burger here. I don't see them *in the present* as inseparable. I think they *should be* inseparable to make their application more apparent, how they relate to the other objects and where the shared overlap of interface/code/feature analysis comes from. I think it is risky to assume people, without our direct help, would be able to discover and remember that onsetslice actually has spectraldiff working under the hood. Furthermore, I don't think naming something spectraldiff versus onsetcurve or noveltycurve vs novelty really opens up a world of possibility in the mind of the user. In some cases it also begs the question as to what the feature actually is. If we say `bufnovelty~` do we mean the novelty represented as the self-similarity matrix, or the _curve_ that is rendered from running the kernel over the SSM.

weefuzzy:
* I agree that `<original domain><something>` is important for anyone to have a hope of relating these back to their slicer counterparts and thus using them for their most obvious purpose, which is to help with segmenting
* Whilst I appreciate the push towards genericness, I'm not sure this is the place: these objects are still tied to audio input (so aren't all that generic), and I don't think there's a helpful way to package up the different change detection mechanisms they use for more general cases (which doesn't mean that I wouldn't one day like non-fft-ing control rate input versions to do Bad Things with). 
* IOW, I'm pretty dead set against `ampdiff` and `spectraldelta` on the grounds that they only *seem* like they have any explanatory power and, as @rconstanzo notes, there's no consistency 

Alternatives to curve that spring to mind: feature; detect

tremblap:
ok that will be a long one. I will read and re-read and try to agree, then I will argue. for me, our agenda of trying to not impose our bias on generic features (all of which they are) and consistency of interface (with transient -> transientslice, and all other time series not having a suffix (curve, feature, etc) and the real-time vs non-real-time aspect which they share a lot more with descriptors than slicers) are all against pigeon-holing them with a single use name binding

so let the brainstorm continue.

novetly is going to be hard to add anything convincingly, although feature is the less worse. for interface consistency, should we do pitchfeatures (plural as it has 2 dimension) ? I am being controversial here to remind interface consistency as the most important part of this discussion.

noveltyfeature at least has the advantage to imply a continuum

onsetfeature is bad. what we do with that time series is to make spectral differences. anything that goes away from this will be a false shortcut I am pretty convinced. spectraldifferencefeature starts to be long

the detrender - ampdiff is actually quite consistent with ampslice and ampgate. ampfeature means nothing, we need to hint at the differential that is happening

so bring along more thoughts, and feel free to quote our own papers at ourselves. I am going for our primary goals (consistency and flexibility and making sure we don't overcurate use) and also our workshop experience (seeing the time series under the hood helped, but required still explanation, which is sexy on the learn platform, but we still had to explain it was the 'green' line that was drawn...)

tremblap:
onsetmetric at least shares a name with the underlying code/algo - with the suffix it becomes onsetmetricfeature which is a great paragraph :)

jamesb93:
> noveltyfeature

`fluid.bufnoveltyfeature~ @features myfeatures @feature 0`

If I have any veto power its over this specific use case arising 😄

weefuzzy:
I think these are all red herrings

> for me, our agenda of trying to not impose our bias on generic features (all of which they are)

In general all we can do is mitigate whatever impositions our choices end up making, not avoid it entirely, especially when there's not a clear best answer like here. Also, in each of these cases, that boat has already sailed: the way the slicers are named already reflects, e.g.
*  a decision to privilege mapping back to a particular section of music technology literature:
    *   `OnsetSlice` is so called because it does what a bunch of 1990s-2000s papers proposed under the banner of onset detection, even though these are strictly tied variously tightly to the assumption that one is in the Fourier domain). 
* a choice to map a concept to a particular underlying model 
    * `Transient[Slice]` models transients as impulsive disturbances to a linear model in a way taken from audio restoration. Other ways of conceiving of what a 'transient' might be are unaccounted for 
* or both
    * `NoveltySlice` is so called because that's the term that Foote used (so mapping back to particular portion of literature) but that is by no means the only documented scheme for devising change detection functions from multivariate input: there are many in the wider time series analysis literature. 'Novelty' is also a hugely culturally loaded term, but we've let that slide in favour of relating it to orthodox MIR practice.   


>  and consistency of interface (with transient -> transientslice,

No – what comes out of `transients` is not at all equivalent to what would come out of a (non-existent) `transient<curve | whatever>` object. For one, you're not seeing the detection function, for another the detection functions between `transients` and `transientslice` don't end up being the same because the former feeds back its interpolated residual to the detector. It's not even yet clear to me that a transients version of these objects would be all that useful, such is the horribleness of the detection inputs and complexity of detection. 

> and all other time series not having a suffix (curve, feature, etc) and the real-time vs non-real-time aspect which they share a lot more with descriptors than slicers) are all against pigeon-holing them with a single use name binding

Well, I can enjoy the inversion of our usual roles in these discussions and say 'won't somebody please think of the novices': treating these things as control signals is (more) advanced usage. I think the thing to privilege here is being able to stick out one's hand and reach for the complementary detection function object for a given slicer, not least because the slicers are comparatively really hard to use without them. 

> novetly is going to be hard to add anything convincingly,

I think you are overestimating the genericness of the current `novelty<word of power>` object. It is still tied to the idea of operating on an already curated (and IMO contentious) menu of FFT-derived features. A more generic version, more worthy of a suffix-less name would just accept any old multivariate input and leave it to the user to compose that how they saw fit (so it would be just making a change detection function based on self-similarity), opening the way to combing different input features, different time frequency representations and so on. This isn't that (and if we ever had that, what would we call it?)

Meanwhile, I'm not sold on names that privilege hinting at implementation details, like `ampdiff`, because I don't think it's possible to be clear (diff what?) or that this is immediately useful information. 

My current preference is for `*Detect` because (to me, once the pattern is established) it makes it perfectly clear that these are the functions that map to the respective slicers. It will still need documenting, but so would anything we pick I think. Meanwhile, exciting affordances like using these as control sources can happily be revealed in docs / examples.

jamesb93:
> I think these are all red herrings
> 
> > for me, our agenda of trying to not impose our bias on generic features (all of which they are)
> 
> In general all we can do is mitigate whatever impositions our choices end up making, not avoid it entirely, especially when there's not a clear best answer like here. Also, in each of these cases, that boat has already sailed: the way the slicers are named already reflects, e.g.
> 
> * a decision to privilege mapping back to a particular section of music technology literature:
>   
>   * `OnsetSlice` is so called because it does what a bunch of 1990s-2000s papers proposed under the banner of onset detection, even though these are strictly tied variously tightly to the assumption that one is in the Fourier domain).
> * a choice to map a concept to a particular underlying model
>   
>   * `Transient[Slice]` models transients as impulsive disturbances to a linear model in a way taken from audio restoration. Other ways of conceiving of what a 'transient' might be are unaccounted for
> * or both
>   
>   * `NoveltySlice` is so called because that's the term that Foote used (so mapping back to particular portion of literature) but that is by no means the only documented scheme for devising change detection functions from multivariate input: there are many in the wider time series analysis literature. 'Novelty' is also a hugely culturally loaded term, but we've let that slide in favour of relating it to orthodox MIR practice.
> 
> > and consistency of interface (with transient -> transientslice,
> 
> No – what comes out of `transients` is not at all equivalent to what would come out of a (non-existent) `transient<curve | whatever>` object. For one, you're not seeing the detection function, for another the detection functions between `transients` and `transientslice` don't end up being the same because the former feeds back its interpolated residual to the detector. It's not even yet clear to me that a transients version of these objects would be all that useful, such is the horribleness of the detection inputs and complexity of detection.
> 
> > and all other time series not having a suffix (curve, feature, etc) and the real-time vs non-real-time aspect which they share a lot more with descriptors than slicers) are all against pigeon-holing them with a single use name binding
> 
> Well, I can enjoy the inversion of our usual roles in these discussions and say 'won't somebody please think of the novices': treating these things as control signals is (more) advanced usage. I think the thing to privilege here is being able to stick out one's hand and reach for the complementary detection function object for a given slicer, not least because the slicers are comparatively really hard to use without them.
> 
> > novetly is going to be hard to add anything convincingly,
> 
> I think you are overestimating the genericness of the current `novelty<word of power>` object. It is still tied to the idea of operating on an already curated (and IMO contentious) menu of FFT-derived features. A more generic version, more worthy of a suffix-less name would just accept any old multivariate input and leave it to the user to compose that how they saw fit (so it would be just making a change detection function based on self-similarity), opening the way to combing different input features, different time frequency representations and so on. This isn't that (and if we ever had that, what would we call it?)
> 
> Meanwhile, I'm not sold on names that privilege hinting at implementation details, like `ampdiff`, because I don't think it's possible to be clear (diff what?) or that this is immediately useful information.
> 
> My current preference is for `*Detect` because (to me, once the pattern is established) it makes it perfectly clear that these are the functions that map to the respective slicers. It will still need documenting, but so would anything we pick I think. Meanwhile, exciting affordances like using these as control sources can happily be revealed in docs / examples.

Agreed. I think a stronger sense of software openness and learning is cultivated by obviating the connection between the code, rather than framing these objects as something which they are not: generic.

tremblap:
thanks @weefuzzy for that thorough answer, as usual. I didn't reply because I am in full agonistic dialogical mode in my head, with all these thoughts fighting each other.

tremblap:
For the last 3 days, I keep swapping words in my head. 'detect' is what we do on the produced time series so it feels one level too abstract. 'curve/stream/flux' imply something that is problematic. 'metric' is boring. then I turn back to the first one and start again.

I think I have started to almost agree with Owen (despite the exaggerations in his arguments) that just the name is maybe misleading the other way. So we're stuck. except if we take detect to imply a continuum in the English language. but then if I am confused, imagine someone who doesn't have the luxury of research time on interface semantics...

AlexHarker:
@jamesb93 asked me to look here, but TLDR (skimmed) so at the risk of repeating someone else - have you considered (for naming):

```
noveltyfeature~
onsetfeature~
transientfeature~
etc.
```

(I get that people here are discussing buffer versions, but if it can run in realtime, why wouldn't that also exist.

[Runs away quickly, suffering the loss of just one spanner]

tremblap:
@AlexHarker buf and RT will exist don't worry

Feature was considered higher in this thread indeed and even was vetoed by @jamesb93

AlexHarker:
I see (from a quick search) that feature has been discussed above:

"onsetfeature is bad"

I disagree - it is the feature on which onset detection is based. Using "spectral difference" means the user needs to know that that's how onsets are found. It also implies a pure difference, but the spectral difference for onset detection is often weighted or biased by time. This is *always* in one direction (you could do the reverse to make a offset detector, but the underlying altos you are providing don't do that). Therefore the spectral differences you are working with are inherently tied to the notion that difference should be weighted in one direction to detect onsets. Detection is poor, because you are removing the detector and giving just the underlying time series - or *feature*.

jamesb93:
I think outside of the one case where I have a sense of dread `fluid.bufnoveltyfeature~ @features mybuf @feature 0`, `feature` as an affix feels quite good.

weefuzzy:
So after some more back and forth `<X>Feature` it is, with the additional side effect that the parameter formerly known as `feature` in `Novelty<Y>` becomes `algorithm`

weefuzzy:
Nope. All in, maximum chaos 🔥

tremblap:
> Excuse my confusion, will the @algorithm change come in this PR or separately and after beta6?

good question - is it used in the videos?

weefuzzy:
Unless you want to put up with `@feature` vs `@features` for any length of time?

jamesb93:
> > Excuse my confusion, will the @Algorithm change come in this PR or separately and after beta6?
> 
> good question - is it used in the videos?

I used onsetslice in the mapping ones so I think we're good on my side of things.

weefuzzy:
Other option is we defer the change to `NoveltySlice` and live with the inconsistency for a bit

jamesb93:
> Unless you want to put up with `@feature` vs `@features` for any length of time?

>:(

jamesb93:
> Other option is we defer the change to `NoveltySlice` and live with the inconsistency for a bit

It doesn't break videos so I think go for it!

tremblap:
if it doesn't break the videos then indeed let's do it earlier than later!

jamesb93:
Docs will appear here:

https://github.com/flucoma/flucoma-docs/pull/101

tedmoore:
I offer you my unsolicited approval 😛

PR #118, opened 2022-03-20
[Fix] Make error language more consistent

jamesb93:
In two cases the use of the word label is out of date with how we currently teach the toolkit. This has been replaced with identifier in the appropriate error message. Similarly `id` should be replaced with identifier as to not confuse native speakers.


weefuzzy (changes requested):
Can you update the function name as well (DuplicateId() or DuplicateIdentifier()). It's called in DataSetClient and LabelSetClient, but you probably already knew that

jamesb93:
> Can you update the function name as well (DuplicateId() or DuplicateIdentifier()). It's called in DataSetClient and LabelSetClient, but you probably already knew that

I expected one might want this :) can do

jamesb93:
> Can you update the function name as well (DuplicateId() or DuplicateIdentifier()). It's called in DataSetClient and LabelSetClient, but you probably already knew that

This is done now.

PR #125, opened 2022-03-25
add dataset version of inverse PCA

weefuzzy:
Complements the point-wise PCA inverse with a batch `DataSet` version, as requested by @tedmoore


tremblap:
will need sc, max, pd, doc PRs too

PR #129, opened 2022-04-13
`NoveltySlice`: Fix reset after parameter change (issue #128)

### LINKED ISSUE #128: noveltyslice misbehaving
if one opens the musical example tab (2) of the max helpfile (correcting the algorithm in the message to get to mfcc) one gets either nothing detected, or a continuous flow of attacks. One then needs to change the threshold to get something sensible out but it is very tempremental.

to reproduce:
- open the help with DAC off, on the 2nd tab
- choose nicol drum loop
- play - nothing or a lot of clicks.
- if nothing, press reset. usually, a lot of clicks
- press the message to change preset. usually a lot of clicks. raising the threshold a little helps to tame but is strangely inconsistent.

this happened first in Pd and is tested behaviour in Max too. Beta5 and beta6. I will download beta4 and see if it is the same.

weefuzzy:
fixes #128 by properly resetting the client's frame offset counter when the window / hop size change. Also zero the internal peak buffer for good measure.

### DIFF
--- a/include/algorithms/public/NoveltySegmentation.hpp
+++ b/include/algorithms/public/NoveltySegmentation.hpp
@@ -33,6 +33,7 @@ public:
   {
     mNovelty.init(kernelSize, filterSize, nDims);
     mDebounceCount = 1;
+    mPeakBuffer.setZero(); 
   }
 
   double processFrame(const RealVectorView input, double threshold,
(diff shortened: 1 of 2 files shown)


tremblap:
I'm checking this now but I need to set up my machine again...

tremblap (approved):
this works thanks. once it is merged I will change the error in the naming in client's verbose string (as per email), and will push to dev the threshold tweak in all 3 cces

PR #133, opened 2022-04-29
NMFClient: Add resynth mode to de-magic interface

weefuzzy:
In anticipation of the buffer management PR, this changes the interface of NMF to add a `resynthMode` signalling whether or not resynthesis should occur. Whilst it introduces verbosity, it reduces the amount of magic in the object. 

Also now allow resynth from fixed bases + activations (which, e.g., allows post processing analysis results and then resynthesizing components). 

Finally, diagnostic messages are updated to use the same terminology as the interface. 

There will be an accompanying PR for SC

### DIFF
--- a/include/clients/nrt/NMFClient.hpp
+++ b/include/clients/nrt/NMFClient.hpp
@@ -39,6 +39,7 @@ enum NMFParamIndex {
   kStartChan,
   kNumChans,
   kResynth,
+  kResynthMode,
   kFilters,
   kFiltersUpdate,
   kEnvelopes,
@@ -55,6 +56,7 @@ constexpr auto BufNMFParams = defineParameters(
     LongParam("startChan", "Start Channel", 0, Min(0)),
     LongParam("numChans", "Number Channels", -1),
     BufferParam("resynth", "Resynthesis Buffer"),
+    LongParam("resynthMode","Resynthesise components", 0,Min(0),Max(1)),
     BufferParam("bases", "Bases Buffer"),
     EnumParam("basesMode", "Bases Buffer Update Mode", 0, "None", "Seed",
               "Fixed"),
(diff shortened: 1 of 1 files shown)


jamesb93:
I didn't fully realise how cool this could be with the new resynth options! Testing coming...

jamesb93:
First thing I've noticed is that the default `resynthMode` is not the old behaviour. My is that it's going to get frustrating constantly changing that attribute when the common usage is to get some resynth from a source buffer given some components.

weefuzzy:
Don't know how you could recreate the old behaviour. Either it's different when the resynth buffer is present and the flag is 0, or it's different when it isn't present and the flag is 1.

jamesb93:
> Also now allow resynth from fixed bases + activations (which, e.g., allows post processing analysis results and then resynthesizing components).

I was experimenting with something like this. It's not particularly useful (though I anticipate some more refined patching might lead somewhere) but it proves that it works.

image

jamesb93:
> Don't know how you could recreate the old behaviour. Either it's different when the resynth buffer is present and the flag is 0, or it's different when it isn't present and the flag is 1.

I guess I just misunderstand the attribute then.

weefuzzy:
> I guess I just misunderstand the attribute then.

It means 'should I try and resynthesize' (at which point the `resynth` buffer will actually be checked to see if anything is assigned, and whether it's actually a buffer etc)

jamesb93:
> 

In that case, is it not the case that the default behaviour was mode 1?

weefuzzy:
No, because it depended on whether `resynth` was assigned. If it was, then this is the equivalent of `resynthMode 1` but otherwise, it would be the equivalent of `resynthMode 0`. There's no way of perfectly replicating that now. The issue being that, with the buffer managment PR, if `resynth` isn't assigned by the user, it will be assigned by the the wrapper so we need some way of being able to not resynthesise. 

i.e. 
`[fluid.bufnmf~ @bases bar @activations foobar]` used to implicitly mean don't resynthesize. But now (with the PR), `@resynth` would be internally given a managed `buffer~` `u127373744` or whatever, so would unconditionally resynthesise.

jamesb93:
> if `resynth` isn't assigned by the user, it will be assigned by the the wrapper so we need some way of being able to not resynthesise.

Right! This makes sense to me now.

PR #136, opened 2022-05-03
Fix/scaler invert

weefuzzy:
Replaces `invert` parameters in scalers with `inverseTransform<Point>` messages


jamesb93 (approved):
Tested both methods and works as expected

tremblap (commented):
tested in Pd (see new 'helpfile' to see in action)

we need to change flucoma-docs though

PR #141, opened 2022-05-08
[Enhance] ChoicesParam for LoudnessClient and PitchClient

jamesb93:
In keeping with the consistency of SpectralShape and BufStats, I think that Loudness should also include a ChoicesParam. This implements it in what I thought was the most similar way to SpectralShape and which is stylistically pretty much the same.

### DIFF
--- a/include/clients/rt/LoudnessClient.hpp
+++ b/include/clients/rt/LoudnessClient.hpp
@@ -25,6 +25,7 @@ namespace client {
 namespace loudness {
 
 enum LoudnessParamIndex {
+  kSelect,
   kKWeighting,
   kTruePeak,
   kWindowSize,
@@ -33,6 +34,7 @@ enum LoudnessParamIndex {
 };
 
 constexpr auto LoudnessParams = defineParameters(
+    ChoicesParam("select","Selection of Outputs","loudness","truepeak"),
     EnumParam("kWeighting", "Apply K-Weighting", 1, "Off", "On"),
     EnumParam("truePeak", "Compute True Peak", 1, "Off", "On"),
     LongParam("windowSize", "Window Size", 1024, UpperLimit<kMaxWindowSize>()),
@@ -61,13 +63,13 @@ public:
   static constexpr auto& getParameterDescriptors() { return LoudnessParams; }
 
   LoudnessClient(ParamSetViewType& p)
-      : mParams(p), mAlgorithm{get<kMaxWindowSize>()}
+      : mParams(p), mAlgorithm{get<kMaxWindowSize>()},
+        mMaxOutputSize{asSigned(get<kSelect>().count())}
   {
     audioChannelsIn(1);
-    controlChannelsOut({1,2});
+    controlChannelsOut({1,mMaxOutputSize});
     setInputLabels({"audio input"});
     setOutputLabels({"loudness and peak amplitude"});
-
     mDescriptors = FluidTensor<double, 1>(2);
   }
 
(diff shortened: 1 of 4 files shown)


tremblap:
I've added myself and Ted here as reviewers as this will need an interface discussion. To unpick:

- if we do it, pitch should have it too, obviously, as the only other non-multivariate descriptor
- in effect, neither of those actually need this interface per se if we stick to their original raison d'être: select was invented to remove an object in the workflow when one or more non-contiguous items were desired (mean and max for instance). The 'bonus' of providing the name of what channel is what was good but just because there was a lot of them... otherwise, we would need MFCC0 and MEL0 and all those guys, which are uselessly complicating the interface.
- I think that not having them in the 101 videos is more pedagogically fruitful than having them. @numchans is a bit harder but yields a lot more potential.

Now, we can discuss it all. if we agree on the last point, we can park this to v1 milestone.

jamesb93:
> * if we do it, pitch should have it too, obviously, as the only other non-multivariate descriptor

I agree. The code is a little more entangled because the client handles scaling and replacing poor measurements (0hz) and I couldn't untangle it myself. It should, IMO, be a separated PR regardless.

> * in effect, neither of those actually need this interface per se if we stick to their original raison d'être: select was invented to remove an object in the workflow when one or more non-contiguous items were desired (mean and max for instance). The 'bonus' of providing the name of what channel is what was good but just because there was a lot of them... otherwise, we would need MFCC0 and MEL0 and all those guys, which are uselessly complicating the interface.

I think that this line of reasoning is needlessly overthinking it. People, including myself and those I have helped in their patching regularly just want one of the things out of these objects and this removes an object to do so and makes the code self documenting. It is unlikely you would decouple several MFCCs that are non-contiguous and separated from each other in a non-programmatic way. There is also the argument of consistency, in that these "semantic" multi-channel outputted descriptors (unlike mel/mfcc/stft) could be made more homogenous in their interface. This is good for all the reasons that are purported in previous discussions related to interface. I am thinking here of the differences between sourceA / source1 between objects, or references to interpolation being named differently.

> * I think that not having them in the 101 videos is more pedagogically fruitful than having them. @numchans is a bit harder but yields a lot more potential.

I've always found `numChans` and `startChan` hand-wavey to be honest. Again, it comes back to the notion of the code being expressive and self-documenting. We want loudness so we write loudness. We want confidence, we write confidence. Sometimes people will care which channel you need to extract from a multivariate descriptor, but in this case it's only ever 2 so enforcing people to care about the layout feels like an imposed difficulty that adds nothing to the usefulness of these objects. Furthermore, a `select` parameter doesn't exclude people using `fluid.bufselect~` or `numChans`. I would advocate any day to give people across a wide range of dispositions options. You like counting arrays and being particular about the layout of data and lots hate it and want to get The Thing :)

tremblap:
I find most of your arguments compelling, although I am not certain that removing programmatic notions of the 101s is a good idea. We can talk about it quickly in tomorrow's meeting to see the best way forward with the various milestones and deadlines.

tremblap:
Yes what you just say there are the arguments I read in the previous answer, and that I find very convincing. You know I love consistency of interface. we could even say I'm obsessed :)

include/clients/rt/LoudnessClient.hpp
```diff
@@ -33,6 +34,7 @@ enum LoudnessParamIndex {
 };
 
 constexpr auto LoudnessParams = defineParameters(
+    ChoicesParam("select","Selection of Outputs","loudness","truepeak"),
```
weefuzzy, line 37: this isn't always true peak, so should probably just say peak

include/clients/rt/LoudnessClient.hpp
```diff
@@ -122,6 +133,8 @@ class LoudnessClient : public FluidBaseClient, public AudioIn, public ControlOut
   algorithm::Loudness                                mAlgorithm;
   BufferedProcess                                    mBufferedProcess;
   FluidTensor<double, 1>                             mDescriptors;
+
+  index mMaxOutputSize;
```
weefuzzy, line 137: this is always 2, so no need for a variable

weefuzzy (changes requested):

include/clients/rt/LoudnessClient.hpp
jamesb93: Indeed - will change

include/clients/rt/LoudnessClient.hpp
jamesb93: Okay cool! Less code is good :)

include/clients/rt/LoudnessClient.hpp
jamesb93: Resolved in 3baac1d

include/clients/rt/LoudnessClient.hpp
jamesb93: Resolved in b75b213

jamesb93:
@weefuzzy can you please implement PitchClient thank you

weefuzzy (approved):

weefuzzy:
This entails changes for SC classes as well : https://github.com/flucoma/flucoma-sc/pull/101

tedmoore:
Should we just merge this in since the other `select` PRs have already been merged?

jamesb93:
> Should we just merge this in since the other `select` PRs have already been merged?

I think so. I'm going to merge the core changes and then the SC classes so the nightlies work.

Issue #151, opened 2022-05-31
Latency reporting for novelty and onset objects may be "incorrect"


AlexHarker:
The use of quotes relates to the fact that we would need to agree on a model of how these control data objects timings should relate to the input stream and it is possible that that model is different between me and others.

In addressing #148 I have suffered some issues regarding alignment of my results (with two objects) and the original novelty feature object:

- *The two processes do not align*. There is an additional delay in the output of the novelty feature object that is equal to window_size / hop_size (so you have an additional set of outputs at the front). That tracks with the object latency reporting (and that of the comparable slicers), which does not attempt to account for the FFT window size (only returning values in multiples of the hop size - in the case of the onset objects I'm not sure I understand why that is, although I get for novelty that you are entering the kernel/filter). In both cases it is possible that there is 1 * hop_size that should really be the window size.

I am not yet 100% sure if I think that the model for latency is incorrect in these objects, but I suspect that it may be. My reading for an onset/novelty or similar is that it should be tagged at the start of the comparable window, but right now I think what will be happening is that it is reported at the end of that window. For Mel bands or mfccs the timing is set to the start of the frame. I can't easily force my results to align for consistency as I don't know the window settings by the time the data reaches my object, but it seems to me that doing mfccs externally and then doing a novelty feature should align with doing it all in one object.

AlexHarker:
Here are two pics of a short file that has been analysed for mfccs and novelty. 

The first is represents that current latency reporting, which to me is noticeably lagged.

<img width="572" alt="Screenshot 2022-05-31 at 11 20 29" src="https://user-images.githubusercontent.com/4799074/171152074-1c7f23a4-3c40-453a-aee2-2c6384fb7bf8.png">

This picture reflects what I think the latency reporting should perhaps be:

<img width="585" alt="Screenshot 2022-05-31 at 11 17 14" src="https://user-images.githubusercontent.com/4799074/171152227-82738c6c-d327-4e41-92d2-085f9e4c388a.png">

AlexHarker:
Another note - the latency of novelty slice and novelty feature are reported as the same, but novelty slice has a peak finder that requires a one hop lookahead, so they should probably (to my mind) be different.

AlexHarker:
OK - I've also now done some testing on the novelty slice object and I'm convinced it is wrong - if I compensate externally for the window size I get much better matches to segments.

[...]

PR #154, opened 2022-05-31
Attempt to correct novelty object latency reporting

AlexHarker:
This is my attempt to correct latency reporting for novelty. My assumptions are:

- That the FFT window needs compensating
- That both objects require compensating for half the filter (rounded up - based on the pre-existing approach) and half the kernel size
- That the slicer requires an additional hop for the peak finder

The previous state was that these were the same in both cases, neither compensating for the FFT window and BOTH correcting for an additional hop that I assume is the hop for the peak finder (which needs to lookahead by one).

I can probably provide audio and a test file to demonstrate why I think this is  better if needed, but I'd need some time to put it together.


AlexHarker:
This relates to #151 .

Issue #171, opened 2022-06-09
CepstrumF0.init() is very expensive (issue for BufPitch)


AlexHarker:
There are a number of reasons this might get called in a buffer client as far as I can tell (setup/reset etc.) and the allocation and trig costs are high. 

In a patch I have that analyses lots of chunks of audio this is the dominant cost of including pitch analysis in the process (so with it the cost is approx 530 secs and without it is approx 115 secs). Any strategies to only call this where absolutely necessary would be good. I'll try to add something where it doesn't get called if you aren't using that particular pitch algorithm and some other stuff as a PR for consideration.

A more complex (but perhaps relevant) observation is that for table type data (DCT/FFT) data can probably be shared between threads, although the model for how clients works may not allow that. It's just frustrating to suffer the cost of setting up multiple tables where it's not strictly needed.

AlexHarker:
Addressed in f6ec9be4ea4af8a3e9b5030db8690a7e2a03a123

PR #172, opened 2022-06-09
Faster pitch init

AlexHarker:
This PR addresses #171.

Essentially it aims not to reallocate memory or recalculate tables where these can be avoided and thus it runs a lot faster in scenarios where BufPitch is being bombarded with small files and thus calling reset() and hence cepstrumF0.init() frequently.

The speed increase for my scenario (in which BufPitch is just one cog) is around 5 times faster, so this cost saving can be really *very* significant.

Hopefully the changes will be obvious - I also note that there is no mInitialised in DCT which seems a bit odd, but probably this is not 100% consistent across the codebase.

### DIFF
--- a/include/algorithms/public/CepstrumF0.hpp
+++ b/include/algorithms/public/CepstrumF0.hpp
@@ -32,8 +32,11 @@ public:
 
   void init(index size)
   {
-    // avoid allocation of maxSize^2 at constructor
-    mDCT = DCT(size, size);
+    // avoid allocation of maxSize^2 at constructor, but don't reallocate if not needed
+    
+    if (mDCT.maxOutputSize() < size || mDCT.maxInputSize() < size)
+      mDCT = DCT(size, size);
+      
     mDCT.init(size, size);
 
     mCepstrum = mCepstrumStorage.segment(0, size);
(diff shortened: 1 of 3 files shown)


tremblap:
@AlexHarker can you see if it is still relevant a PR or close it if not? thanks

AlexHarker:
I believe that some equivalent changes have been made (in that there won't be a reallocation when DCT:init() gets called. However, I suspect that the speed-up here was coming from not reinitialising the DCT at all within CepstrumF0::init() if the DCT table sizes aren't changing - that change in this PR isn't part of the changes that have gone into the main branch.

Thoughts @weefuzzy? Perhaps you don't like this kind of pattern, but it could also be an early exit in DCT::init() which would also potentially benefit other situations. I can do that as a PR (or here) if desirable.

weefuzzy:
I *think* my reason for not spending much time on this before was that I was more inclined to stop doing the DCT this N**2  way altogether, and do the DCT via an FFT instead for `log`n goodness. 

Given that I have no bandwidth for that now, I'm happy enough to improve the existing scheme. Ideally the usage pattern would be similar the FFT objects, but that might  be hard.

AlexHarker:
I am closing this as it is superseded by #241 and this will be linked for relevant conversation if needed.

PR #190, opened 2022-07-19
WIP: Making Real-time objects allocator-aware

weefuzzy:
Work in progress on making the real-time clients allocator aware, and then real-time safe. 

* I've tried to preserve the original form as much as possible, so the Eigen code should broadly read the same, except for using `Eigen::Map` in place of Eigen's allocating  containers. This is encapsulated in a wrapper template `ScopedEigenMap` which takes care of setting up the storage (with `std::vector`) and maintaining a `Eigen::Map`. We also have aliases for STL containers needed for RT use that use the right allocator. 
*  OTOH, `MedianFilter`'s `deque`-based design didn't play well with the SC memory pool at all, so it's been changed to do the same thing but with vector instead. I think we should revisit the double-heap algorithm for this in the future. 
* `FFT` now has a static `FFTSetup` to draw upon, which saves a great deal of duplicate computation and non-R-safe allocation, at the cost of imposing a globally maximum FFT size (currently 2^16). 
*  I'm using https://github.com/foonathan/memory to deal with the allocator framework. The equivalent standard library stuff isn't universally deployed yet, and the boost equivalent looks a bit of a mission in terms of getting set up. This seems to work. 

### Checklist
- [x] AmpFeature
- [x] AmpGate
- [x] AmpSlice
- [x] AudioTransport
- [x] Chroma
- [x] HPSS
- [x] Loudness
- [x] MFCC
- [x] MelBands
- [x] NMFFilter
- [x] NMFMatch
- [x] NMFMorph
- [x] NoveltyFeature
- [x] NoveltySlice
- [x] OnsetFeature
- [x] OnsetSlice
- [x] Pitch
- [x] Sines
- [x] SpectralShape
- [x] Transient
- [x] TransientSlice
- [x]  Fix tests (`TransientSlice` behaviour changed due to bug fix – drums tests don't pass)
- [x] Fix NRT wrapped clients
- [x] (S)KMeans
- [x] Scalers 
- [x] KDTree
- [x] MLP
- [x] KNNClassifier/Regressor
- [x] UMAP 
- [x] Parameter types 😬


weefuzzy:
CI checks aren't going to work for a while while stuff churns

PR #213, opened 2023-01-27
added-merge-to-labelset

tremblap:


weefuzzy (approved):
Aside from a couple of formatting weirdness-es, this looks fine. 

Longer term, this algorithm probably belongs in `FluidDataSet` itself, and can perhaps be re-spelled more readably using standard library algorithms.

PR #216, opened 2023-02-16
better behaving real time query of KDTree

tremblap:
In this first version, I fixed when (k * pointsize) > outbuf.numFrames - now it fills up to the size of the outbuffer. 

We discussed outputing the size of the number of valid points, thus opening the door to using the radius in RT (always up to the size of the outbuffer) but I have questions on the code.

We can also merge this and keep the options for later.

### DIFF
--- a/include/clients/nrt/KDTreeClient.hpp
+++ b/include/clients/nrt/KDTreeClient.hpp
@@ -188,7 +188,7 @@ public:
       auto kdtreeptr = get<kTree>().get().lock();
       if (!kdtreeptr)
       {
-        // c.reportError("No FluidKDTree found");
+        // c.reportError("No FluidKDTree found"); //why are both this and line 197+214 commented?
         return;
       }
 
@@ -206,7 +206,6 @@ public:
                                 get<kOutputBuffer>().get()))
         return;
       auto datasetClientPtr = get<kDataSet>().get().lock();
-      // if (!datasetClientPtr) datasetClientPtr = mDataSetClient.get().lock();
       if (!datasetClientPtr)
         datasetClientPtr = kdtreeptr->getDataSet().get().lock();
 
@@ -219,8 +218,9 @@ public:
       auto  dataset = datasetClientPtr->getDataSet();
       index pointSize = dataset.pointSize();
       auto  outBuf = BufferAdaptor::Access(get<kOutputBuffer>().get());
-      index outputSize = k * pointSize;
-      if (outBuf.samps(0).size() < outputSize) return;
+      index realK = std::min(k, (outBuf.samps(0).size() / pointSize));
+      if (realK <= 0) return;
+      index outputSize =  realK * pointSize;
 
       RealVector point(dims, c.allocator());
       point <<= BufferAdaptor::ReadAccess(get<kInputBuffer>().get())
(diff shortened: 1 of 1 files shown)


tremblap:
test code:
``` Supercollider
a = Dictionary.newFrom([\cols, 3, \data, Dictionary.newFrom(4.collect{|i|[i.asSymbol, i.dup(3)]}.flatten)])

d = FluidDataSet(s)
d.load(a)
d.print

b = Buffer.alloc(s,6)
c = Buffer.alloc(s,3)

k = FluidKDTree(s)
k.fit(d)

x= {arg nb = 1; var poke = Impulse.kr(1); k.kr(poke,c,b,nb); FluidBufToKr.kr(b).poll(poke); Silent.ar();}.play

c.sendCollection(0.dup(3))

x.set(\nb, 4)
```

include/clients/nrt/KDTreeClient.hpp
```diff
@@ -219,8 +218,9 @@ class KDTreeQuery : public FluidBaseClient, ControlIn, ControlOut
       auto  dataset = datasetClientPtr->getDataSet();
       index pointSize = dataset.pointSize();
       auto  outBuf = BufferAdaptor::Access(get<kOutputBuffer>().get());
-      index outputSize = k * pointSize;
-      if (outBuf.samps(0).size() < outputSize) return;
+      index realK = std::min(k, (outBuf.samps(0).size() / pointSize));
+      if (realK <= 0) return;
+      index outputSize =  realK * pointSize;
```
weefuzzy, line 223: I don't get this bit. Surely the only way to know the actual output size is to do the actual query?

include/clients/nrt/KDTreeClient.hpp
```diff
@@ -188,7 +188,7 @@ class KDTreeQuery : public FluidBaseClient, ControlIn, ControlOut
       auto kdtreeptr = get<kTree>().get().lock();
       if (!kdtreeptr)
       {
-        // c.reportError("No FluidKDTree found");
+        // c.reportError("No FluidKDTree found"); //why are both this and line 197+214 commented?
```
weefuzzy, line 191: Because we'd very much like to be able to uncomment it one day and report errors from the RT objects.

include/clients/nrt/KDTreeClient.hpp
```diff
@@ -232,9 +232,9 @@ class KDTreeQuery : public FluidBaseClient, ControlIn, ControlOut
       }
 
       auto [dists, ids] =
-          kdtreeptr->algorithm().kNearest(point, k, 0, c.allocator());
+          kdtreeptr->algorithm().kNearest(point, realK, 0, c.allocator()); // i'd like to pass get<kRadius>() and output min(nbpoints,realK) somehow
```
weefuzzy, line 235: Pass the radius as the 3rd argument. Just pass `k` as is, unless I'm missing something subtle. You can find out how many you actually got from the results.

include/clients/nrt/KDTreeClient.hpp
weefuzzy: Oh, ok, I see that you're making sure that the user's buffer is fit for purpose, so you can bail if not able to take any points. Is the idea to be more permissive than the existing check? I don't `realK` is the right name, IAC, and I'm in two minds about whether it's worth doing this pre-query.

weefuzzy (approved):

PR #227, opened 2023-02-23
FluidDataSet: Remove  mutable, deal with fallout

### LINKED ISSUE #225: `FluidDataSet` has mutable members
It's not entirely clear why, except that the accessors for ids, data etc. are `const` member functions returning mutable views (viz. `FluidTensorView<T,N>` rather than `FluidTensorView<const T, N>`. 

If we really need mutable accessors like this (do we??), then there should be separate `const` and non-`const` overloads rather than using `mutable` without a protecting lock.

Introducing `const` here (as ever) kicks off a small cascade of side effects where callers are (erroneously) assuming mutable return values, so mopping all this up isn't a super-quick job, but should be done.

weefuzzy:
fixes #225

Introduces  `Input<X><Y>View>` aliases into TensorTypes.hpp for easier expression of const views

### DIFF
--- a/include/algorithms/public/DataSetQuery.hpp
+++ b/include/algorithms/public/DataSetQuery.hpp
@@ -36,7 +36,7 @@ public:
     index  comparison;
     double value;
 
-    bool test(RealVectorView point)
+    bool test(InputRealVectorView point)
     {
       using namespace std;
       switch (comparison)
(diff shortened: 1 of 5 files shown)


tremblap (approved):
tested on SC UTs on all 3 OSes

PR #229, opened 2023-02-27
skeleton of voiceallocator, to troubleshoot multi control in management

tremblap:
see the Max repo for the first test.


tremblap:
redundant with #249

PR #235, opened 2023-08-04
EOL `.gitattributes` normalisation

lewardo:
EOL standardisation with `.gitattributes` to fix issues with versions of Git before 2.10 allowed automatic EOL conversion.


jamesb93:
`.gitattributes` look good to me. @weefuzzy I don't know much about `clang-format`. If it looks good to you on both fronts shall we merge?

tremblap:
Thanks for this. @weefuzzy is the clang-format originator but as this is being discussed on another of those (on the max repo one) and once I get the heads up I'll merge them all I think... the checks are being done and working (although I wonder how it could not :) )

weefuzzy:
`.clang-format` is YAMLish, should be grand.

PR #237, opened 2023-08-11
Update issue templates

tremblap:


tremblap:
a proposed template for bug and unexpected behaviour

weefuzzy (approved):
Looks fine to get going with. Thanks.

jamesb93:
I would use the forms feature to make this simpler and more readable.

jamesb93 (approved):

weefuzzy:
Thanks for that, but I think we've lost useful flexibility because now there is nowhere for people to include code snippets (or attachments?), which I think is to be encouraged, welcomed, cherished etc. Personally I'm satisfied with something that encourages useful reports but stops short of trying to mandate exactly what they should look like (it's not like we're dealing with thousands)

tremblap:
@weefuzzy the box allows drag and drop, is that ok or do you have another preferred format than this?

weefuzzy:
Yeah, all good. @jamesb93 explained how computers work for me, and has graciously emboldened the text about drag n' drop for the hard of understanding.

Issue #236, opened 2023-08-11
SineExtraction::init() doesn't reset mBuf


AlexHarker:
The contents of mBuf should have a consistent size, but if it isn't reset when init() is called this can not be the case (leading to possible crashes or undefined behaviour).

When this object was updated the reset was commented out, but it is needed, so I would propose that:

https://github.com/flucoma/flucoma-core/blob/2681d58fd4642c67a4135a28b7e19b0a5f9d5972/include/algorithms/public/SineExtraction.hpp#L57

Should be:

`   mBuf = makeEmptyQueue(alloc);
`

Doing so solves an eigen_assert I get with debug builds. @weefuzzy ?

weefuzzy:
Yes. Or just call `clear()`? Horses for courses.

AlexHarker:
oh sorry - on mBuf - I don't mind either way but it asserts without something and is liable to crash, so it would be good to fix. @weefuzzy happy to PR whichever you'd prefer.

weefuzzy:
Cool, thanks. Very slight aesthetic preference to calling `clear()`.

AlexHarker:
Turns out that clear() is not a thing for a queue.

weefuzzy:
Well then, that makes our choice.

AlexHarker:
Fixed in 7247a3614b60d8e363f670482ef2aa50d7f61b1c

PR #238, opened 2023-08-11
eigen out-of-bounds crash fix

lewardo:
@AlexHarker implement fix discussed in #236 that does indeed fix crash, including those mentioned in flucoma/flucoma-max#386


AlexHarker:
Fixes #236

PR #239, opened 2023-08-11
Appropriate client methods are now const

AlexHarker:
This is desirable as it allows differentiation outside of the client / parameter set of what will touch object state (which applying constraints does not).

This will aid in reasoning about threading things in other places.

### DIFF
--- a/include/algorithms/public/Envelope.hpp
+++ b/include/algorithms/public/Envelope.hpp
@@ -59,7 +59,7 @@ public:
     return fast - slow;
   }
 
-  bool initialized() { return mInitialized; }
+  bool initialized() const { return mInitialized; }
 
 private:
   void initFilters(double cutoff)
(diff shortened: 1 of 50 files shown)


weefuzzy (approved):
Seems cromulent, thanks.

AlexHarker:
Thanks @weefuzzy - are you happy with the further additions before I merge? The most salient one is the enable_if_t refactor (I couldn't get the method const any other way). This was suggested by @lewardo as an alternative incantation, but I cannot say I ever really feel I understand the details of std::enable_if_t

weefuzzy:
Should be equivalent 😬 I think the difference is in the context in which the method is enabled / disabled, so that in the first case it was too 'late' to be consty. Or something.

AlexHarker:
@tremblap - these changes are pretty minimal (there are a lot, but they shouldn't do anything and if they did the compiler should test them) but given the number can you give them a spin in your preferred methods before merging?

PR #243, opened 2023-08-16
Adapt YINFFT to store its FFT, rather than allocating per frame (usin…

AlexHarker:
…g the default allocator)

This is an update for #169.

This is not only a speed optimisation. It also fixes the fact that even in SC it will currently allocate an FFT using the default allocator for every frame whoever yinFFT is called.

I will check the tests, which failed last time and make any necessary updates as a second step.


AlexHarker:
If @weefuzzy has a chance to look at the sizes for instantiation that would be useful.

tremblap:
if it is approved it means it passes my QC. I reproduce expected results and it is faster.

PR #246, opened 2023-08-18
Rudimentary `fluid.polynomialregressor` object

lewardo:
Simple N-to-N polynomial regressor, N parallel monovariate polynomial regressors.  
 - `DataClient`-inherited member functions all work, including `print`, `read`, `write` etc.
 - same interface as `fluid.MLPRegressor`, with `fit`, `predictpoint`, `predict` etc. iface
 - `regressors` parameter sets input and output dataset size, namely the number of parallel regressors
 - `degree` parameter sets the degree of the polynomial regression, currently all regressors are set to the same degree at once

@weefuzzy Potential for discussion on and rethinking of the implemented regression algorithm e.g. Gaussian regression may be more appropriate instead of simple polynomial regression, this is just a proof of concept and initial posit for a deterministic alternative to the artificial neural network objects.

tested in max against lewardo/flucoma-max@02d3b22

_current issues:_ 
 - when `read`ing from json, the parameters update in the algorithm but not in the client so you have to set the parameters (degree, number of regressors) correctly before `read`ing for it to be recognised by the client

patch used to test new object
<pre><code>
1806.3oc2assaaiCD84Tf9OPXzBjtqq.ueI+JKVTHayjnt1RBRxsIaw1u8kj
RJwwlVlpkVtMEAwUzzdl4vCOb3Plu812b0rEEOnqmAtA7Wfqt5alVtx0lskq
5a3pYaReX45zZWGmsrXyFcdyr4cuYi9gF2abM6CIfk2mlemFrJsIsV2.pzqS
axJxquOq7oOw5rb8xhs4tOFtu0xJcs46008O4uK4a2TrsYstw4Iv9lyV4rew
hO+QJ9IqTl1r79r769TkdYSaHREpDnfBUHk.pvPFmLGfj3DNR.wPEmxwXpRZ
ZjASfyADZBD726X9r7dqibM9eu8M1WMuLOB.nv.fU5zUfaqJ1.9bcQ9rQD5n
gBckhjvPRFFyfPkxDzHSTZhOIhZ9+BLDgIXE5oPGCmxPmNGbsxD82Y3Ll3rb
aCXSZYoIFpmEWBBbHTRJkGhRhVPBwULjv7CTY.IkLw1Fky3JAVJYnoFxtlXv
quVk0nAMEikrPTCRVvXOjEDJgRPRhPg3DAECEWHxB1PVjlf+1rFmPyXhawPw
s.ScipLLU.YRIWR8O7KQIRK.QgRlArvzy0neZSS01riDfn9laaq4wRcabLa1
yNRYZU5Fcit5S57zEq0u.Yre8N6TouyL6ptn54oa6hZrwKsvUIT6+3JrQlQx
D3mYK3ojsXblq4F1x0U5ObmNWWk17z5S0ig5PGBDvbbBFgMKgvfTESfTJG0A
Z0JnsPAVIEleCQheyYKqrrEsOlBlLZlBFR7nuRkmMlxFCQO8N8gLEmX5OC5s
KaAMdAVeqFSnm.GvwGGrofDIX.O75LLYnv.4rAC45uZ7zCQA8CkUf2cKB7Gc
+FAgwBUFjbPwrDAyFslWwbqvAkRjTAm2tJ74ZhwoPhHE8nAyQGIPsQOhaCTF
d50AJqx1Ygjexg5AS3fK3IXdHBhD7jOAnAr.r3Hv.9nvvBy9+lMu6U+PxfYS
H4POZBX0O.j7iSMVrso4n4SiFN1CbwzcAD5nADeIkhaWpf96OdLn.gDJt33w
wlxb65sYqR5Rt76.yNVWM5IP1IOGYhCb7ZILlmjqfxKgnpdU1xFPV9PvxnWK
QM5oNdkWkrIWdskqTVr9w7hMYoqeZKXeGrPmlWGQhCYzym7yaNIJ8K9hwngS
JmJBjsbFWLdR1bhbvDvb4cQX79BfN0QZuRQLkIDidF.Ax8ssT0ziG1xbYTMM
gZjPC13kC7MKfglbMyZmZvs.cdS0ie78Y2.d+slehEKYPcRBRztwDFMwHORP
JybDC9PHLaBphS.FjnCF1IIIkoK+GW0rtArrXc8M3QiE1ulrh7zpGOBpPCAU
HXWo8P3omT3vg6pJ1VF+XmEDiP5TI47oe0w0Eoq.OGEf2gmCh5Jl7fF7aWr.
cApRyYVQ.N3VQ3jNEgNt+qVM.AJ.X30oDf.FBC30qB.kDxP+kS.Xusbe7rj9
Q13DBFh7GE5xJDAgSdoY2K5yxiXvqBXfuKzUjKVQo6JO+eBhU4oGbqBjtpS6
+RibFI.SaE3FbUeL1MlKdp7ZSbQ1h7wQL7oQzmh2Y+zHNVzt8eyLdA3XG8DI
rBwad0tjfeQt.FrQP237IO+9802a6l6Fes+0LyYL6arGpTWrsZYePz4e.7y1
akttIK2cKf1sSlgQvNd0lrUkElPtt+34UIr1qsi8hcfITHmq3D6kRPhL6my2
6872VQ0Jc0IxfK73YOW8HAjM0SabGAKZu+Fm1hjAQPi3lQj+fClQgOrv.Je2
hpv6IlB8ciaDGZcO8LJ3ElbtvKLEGbrgOKwVDnBg5uSbjYuxE+rglxVZ2Cnt
ROm9nv6sdJ3dd4I4AwwsENEfFISP4oXg1hqGROiy7WUDXBboGl.yy3qh6oho
inmuJYWnHLDHEveE0YPwPmg8KoDpLjHSMblVLahVDkKeYAwkZk6IOYSglJW1
tO9P8YBVLnSCifSKBhBQ63PwvfvPLXuaECCFzvBoc0kHXOdP1KZgGkLhYJwX
EsfvSLLdFLn7D22sBQ.26pXbIKQvOcOiWrgNIYYXcNojDXrEJJDmXiFz3FKd
Dkf1MaDsGJXkk3r44Qra83Hkg3iYIsXXvflqinwyfAQR6wgXXvv1rCavY78I
FfkjcySv8Xb7RXPbMwvIyzkyUmWt6SwwICQ6Tbp7s59CiwcOD4vCd.K1M8Q2
SmozGCp3hhfpsH2E.j9xIZe7R5z6KT52qwHW8coDVKTae7LkzaXdcvkws8.2
epxsdbZ5TQ2OgvwNzcImIYDNTvYT2eRbDH7TuGVndgdi3BODgHAQrHhW301G
8303IyqgigZ0o7bLlEy+wbjVV9EcUcmgac3YaR+bg6CIm29bVd6yr1mqzeIq
+iz0TZ0x6yZzKa1V4hyYOv6tDVy1TXbg7sY8RIVHyZZ2AIkmtQWWl1BNtCb5
suwzg+G.nuC+O
</code></pre>

### DIFF
--- a/include/algorithms/public/PolynomialRegressor.hpp
+++ b/include/algorithms/public/PolynomialRegressor.hpp
@@ -96,6 +96,15 @@ private:
 
     void setInputSpace(Eigen::Ref<VectorXd> in) const 
     {
+        VectorXd designColumn = VectorXd::Ones(in.size());
+        ArrayXd inArray = in.array();
+        mDesignMatrix.conservativeResize(in.size(), mDegree + 1);
+
+        for(index i = 0; i < mDegree + 1; ++i) {
+            mDesignMatrix.col(i) = designColumn;
+            designColumn = designColumn.array() * inArray;
+        }
+
         mIn = in;
         mInSet = true;
     };
@@ -108,7 +117,7 @@ private:
 
     void getRegressionCoefficients(Eigen::Ref<VectorXd> coefficientsOut) 
     {
-
+        
     };
 
     index mDegree       {2};
--- a/include/algorithms/public/PolynomialRegressor.hpp
+++ b/include/algorithms/public/PolynomialRegressor.hpp
@@ -29,70 +29,84 @@ public:
     explicit PolynomialRegressor() = default;
     ~PolynomialRegressor() = default;
 
-    void init(index degree = 2)
+    void init(index degree, index dim)
     {
         mInitialized = true;
         mDegree = degree;
+        mSize = dim;
     };
 
-    index size() const { return asSigned(mDegree); };
     index dims() const { return asSigned(mDegree); };
+    index size() const { return asSigned(mSize); };
 
     void clear() { mRegressed = false; }
 
-    index   getDegree()     const { return asSigned(mDegree); };
     bool    regressed()     const { return mRegressed; };
     bool    initialized()   const { return mInitialized; };
 
     void setDegree(index degree) {
-        if (mDegree == degree)
-        return;
+        if (mDegree == degree) return;
 
         mDegree = degree;
+        mCoefficients.conservativeResize(mDegree + 1, mSize);
         mRegressed = false;
     }
 
-    void calculateRegressionCoefficients(InputRealVectorView in, 
-                                         InputRealVectorView out,
+    void setSize(index dim) {
+        if (mSize == dim) return;
+
+        mSize = dim;
+        mCoefficients.conservativeResize(mDegree + 1, mSize);
+        mRegressed = false;
+    }
+
+    void calculateRegressionCoefficients(InputRealMatrixView in, 
+                                         InputRealMatrixView out,
                                          Allocator& alloc = FluidDefaultAllocator())
     {
         using namespace _impl;
 
-        ScopedEigenMap<Eigen::VectorXd> input(in.size(), alloc), 
-          output(out.size(), alloc);
+        ScopedEigenMap<Eigen::MatrixXd> input(in.rows(), in.cols(), alloc), 
+          output(out.rows(), out.cols(), alloc);
         input = asEigen<Eigen::Array>(in);
         output = asEigen<Eigen::Array>(out);
 
-        generateDesignMatrix(input);
+        for(index i = 0; i < mSize; ++i)
+        {
+            generateDesignMatrix(input.col(i));
 
-        Eigen::MatrixXd transposeProduct = mDesignMatrix.transpose() * mDesignMatrix;
-        mCoefficients = transposeProduct.inverse() * mDesignMatrix.transpose() * output;
+            Eigen::MatrixXd transposeProduct = mDesignMatrix.transpose() * mDesignMatrix;
+            mCoefficients.col(i) = transposeProduct.inverse() * mDesignMatrix.transpose() * output.col(i);
+        }
+        
 
         mRegressed = true;
     };
 
-    void getCoefficients(RealVectorView coefficients) const
+    void getCoefficients(RealMatrixView coefficients) const
     {
        _impl::asEigen<Eigen::Array>(coefficients) = mCoefficients;   
     };
 
-    void setCoefficients(InputRealVectorView coefficients)
+    void setCoefficients(InputRealMatrixView coefficients)
     {
-        setDegree(coefficients.size() - 1);
+        setDegree(coefficients.rows() - 1);
+        setSize(coefficients.cols());
+
         mCoefficients = _impl::asEigen<Eigen::Array>(coefficients);
         mRegressed = true;
     }
 
-    void getMappedSpace(InputRealVectorView in, 
-                        RealVectorView out, 
+    void getMappedSpace(InputRealMatrixView in, 
+                        RealMatrixView out, 
                         Allocator& alloc = FluidDefaultAllocator()) const
     {
         using namespace _impl;
 
-        ScopedEigenMap<Eigen::VectorXd> input(in.size(), alloc),
-          output(out.size(), alloc);
+        ScopedEigenMap<Eigen::MatrixXd> input(in.rows(), in.cols(), alloc),
+          output(out.rows(), out.cols(), alloc);
         input = asEigen<Eigen::Array>(in);
-        output = asEigen<Eigen::Array>(in);
+        output = asEigen<Eigen::Array>(out);
 
         calculateMappings(input, output);
 
@@ -100,16 +114,20 @@ public:
     }
 
 private:
-    void calculateMappings(Eigen::Ref<Eigen::VectorXd> in, Eigen::Ref<Eigen::VectorXd> out) const
+    void calculateMappings(Eigen::Ref<Eigen::MatrixXd> in, Eigen::Ref<Eigen::MatrixXd> out) const
     {
-        generateDesignMatrix(in);
-        out = mDesignMatrix * mCoefficients;
+        for(index i = 0; i < mSize; ++i)
+        {
+            generateDesignMatrix(in.col(i));
+            out.col(i) = mDesignMatrix * mCoefficients.col(i);
+        }
     }
 
     void generateDesignMatrix(Eigen::Ref<Eigen::VectorXd> in) const
     {
         Eigen::VectorXd designColumn = Eigen::VectorXd::Ones(in.size());
         Eigen::ArrayXd inArray = in.array();
+
         mDesignMatrix.conservativeResize(in.size(), mDegree + 1);
 
         for(index i = 0; i < mDegree + 1; ++i, designColumn = designColumn.array() * inArray) 
@@ -117,11 +135,12 @@ private:
     }
 
     index mDegree       {2};
+    index mSize          {2};
     bool  mRegressed    {false};
     bool  mInitialized  {false};
 
     mutable Eigen::MatrixXd mDesignMatrix;
-    Eigen::VectorXd mCoefficients;
+    Eigen::MatrixXd mCoefficients;
 
 };
 
--- a/include/algorithms/public/PolynomialRegressor.hpp
+++ b/include/algorithms/public/PolynomialRegressor.hpp
@@ -34,13 +34,10 @@ public:
         mDegree = degree;
     };
 
-    index size() const { return asSigned(mInSet ? mIn.size() : 0); };
-    index dims() const { return size(); };
+    index size() const { return asSigned(mInSet ? mIn.size() : 1); };
+    index dims() const { return asSigned(mInSet ? mIn.size() : 0); };
 
-    void clear() 
-    {
-        mInSet = mOutSet = mRegressed = false;
-    }
+    void clear() { mInSet = mOutSet = mRegressed = false; }
 
     index   getDegree()     const { return asSigned(mDegree); };
     bool    regressed()     const { return mRegressed; };
@@ -51,7 +48,7 @@ public:
         return;
 
         mDegree = degree;
-        resetMappingSpace();
+        clear();
     }
 
     void process(RealVectorView in, RealVectorView out)
@@ -71,6 +68,13 @@ public:
        assert(mRegressed);
        _impl::asEigen<Eigen::Array>(coefficients) = mCoefficients;   
     };
+
+    void setCoefficients(InputRealVectorView coefficients)
+    {
+        mCoefficients = _impl::asEigen<Eigen::Array>(coefficients);
+        setDegree(coefficients.size() - 1);
+        mRegressed = true;
+    }
 
     void getMappedSpace(InputRealVectorView in, 
                         RealVectorView out, 
@@ -103,8 +107,6 @@ public:
         setInputSpace(input);
         setOutputSpace(output);
     };
-
-    void resetMappingSpace() { mInSet = mOutSet = mRegressed = false; };
 
 private:
     void setInputSpace(Eigen::Ref<Eigen::VectorXd> in) 
@@ -160,7 +162,7 @@ private:
 
     Eigen::VectorXd mCoefficients;
 
-};   
+};
 
 } // namespace algorithm
 } // namespace fluid
--- a/include/algorithms/public/PolynomialRegressor.hpp
+++ b/include/algorithms/public/PolynomialRegressor.hpp
@@ -52,21 +52,22 @@ public:
         resetMappingSpace(); 
     }
 
-    void process(RealVectorView in, RealVectorView out) {
+    void process(RealVectorView in, RealVectorView out)
+    {
         setMappingSpace(in, out);
         process();
     };
 
     void process() 
     {
-        assert(mInSet && mOutSet && "input and output mapping must be set");
-        
+        assert(mInSet && mOutSet);
+        calculateRegressionCoefficients();
     };
 
     void getCoefficients(RealVectorView coefficients) const
     {
-        VectorXd output;
-        
+        assert(mRegressed);
+        asEigen<Vector>(coefficients) = mCoefficients;   
     };
 
     void setMappingSpace(RealVectorView in, RealVectorView out) const
@@ -90,7 +91,7 @@ public:
         setInputSpace(output);
     };
 
-    void resetMappingSpace() const { mInSet = mOutSet = false; };
+    void resetMappingSpace() const { mInSet = mOutSet = mRegressed = false; };
 
 private:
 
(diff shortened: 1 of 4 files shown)


include/algorithms/public/PolynomialRegressor.hpp
```diff
@@ -0,0 +1,149 @@
...
+    void setCoefficients(InputRealMatrixView coefficients)
+    {
+        setDegree(coefficients.rows() - 1);
+        setDims(coefficients.cols());
+
+        mCoefficients = _impl::asEigen<Eigen::Array>(coefficients);
+        mRegressed = true;
+    }
+
+    void getMappedSpace(InputRealMatrixView in,
```
weefuzzy, line 101: Is this the main processing function? If so, can it be called `process` in line with other algos?

weefuzzy (changes requested):
Thanks @lewardo and congrats on pushing your way through a complete implementation of algo + client, especially with all the boilerplate needed for the dataset stuff. I've not looked in huge detail over the code, but just have general thoughts. 

I've warmed a bit to the idea of this object since it was first suggested. I still don't think that higher-degree regressions are (often) all that useful (because of the overfitting), but certainly having a linear regressor is a generally handy thing. 

We can think / dream about gaussian process regressors in the future. They're much more complex to implement (we'll need a sampling framework, for one thing) and open all kinds of interface questions (e.g. they produce generative models, which we don't have any precedent for yet). 

The problem with refreshing the client's hyperparameters when json is read is meant to have been fixed for the MLPs, so perhaps check the clients for the MLP regressor to see what's done there. 

Meanwhile, other thoughts: 

* if I ever get round to implementing an echo state network for flucoma, we'll need a linear fitter for that as well, except with Tikhonov regularization added ('ridge regression'). That might also be useful for this object, to mitigate overfitting. Thoughts? 
* Can you take me through why we need the `regressors` parameter in the client, rather than just inferring this from the dimensionality of the input dataset? 
* A UX challenge will be trying to know which `degree` works best for the data in question (and we have parallel challenges for the MLPs, in terms of making principled comparisons between alternative models). There's a few ways of approaching this, but I'm wondering if a separate object that implements something contemporary like Pareto Smoothed Importance Sampling (PSIS) or the Widely Applicable Information Critereon (WAIC) would be interesting / feasible. And how, regardless of algorithm, the interface for a model comparison object could work. If you have opinions / interest in looking into that...

include/algorithms/public/PolynomialRegressor.hpp
lewardo: it is indeed the main process, I've renamed it to `process` in lewardo/flucoma-core@7b92c534 in line with the other algorithms, as well as renaming `calculateRegressionCoefficients` to `regress` for brevity

lewardo:
I've implemented some of the changes you mentioned, and am working on various others as of writing  
Changes I've implemented:
 - the `regressors` parameter is no longer, it as been axed in favour of automatic setting, when saving or fitting the object
 - the client parameters now update to reflect the algorithm's state on `load` and `read` messages


I will look into a b-spline regressor to implement as well as this polynomial regressor, as it seems to be more appropriate and have less statistical issues when dealing with noisy data, but I'm also looking into the ridge regression as you suggested as it also seems to be an interesting path to follow

lewardo:
After a discussion with @tremblap we conjured up a potential UI for 1-N rather than as well as N-N regressors - a simple repeating of the 1 input N times, which could be interesting for the user if they should want to map the same domain to many ranges, let me know what you think of this and I could add it in

weefuzzy:
> I've implemented some of the changes you mentioned, and am working on various others as of writing
Changes I've implemented:

Many thanks. Those changes would seem to bring the object up to readiness, save for all the documentation, testing, help files etc etc etc. If you wanted to be ultra virtuous (and postpone some of the foregoing), you could knock up some C++ tests for this algorithm and add them to the `flucoma-core` test suite (which, IIRC, would be the first algo to receive such hallowed treatment). 

FWIW, adding regularisation for ridge regression should be a very simple change (addition of a single scalar parameter, and an extra term in the linear algebra), at least in the linear case. Anyway, it can be left till later.

weefuzzy:
> After a discussion with @tremblap we conjured up a potential UI for 1-N rather than as well as N-N regressors - a simple repeating of the 1 input N times, which could be interesting for the user if they should want to map the same domain to many ranges, let me know what you think of this and I could add it in

Sounds like Surprising Behaviour to me. Would rather a more explicit workflow with less magic (i.e. user duplicates cols of dataset n times if that's what they want)

lewardo:
I've implemented Tikhonov normalisation into the regressor, currently the Tikhonov matrix can only be $\Gamma$ = $\alpha$ $I$ making the linear factor $\alpha^2$ $I$ (as in ridge normalisation), it is however now very easy to change and/or add different filter matricies like Fourier filters etc.

AlexHarker:
It might be relevant to have help from / pass over some of the helpfile work to @MattS6464 given that he is very knowledgeable now in that area...

lewardo:
That would be very helpful if @MattS6464 wanted to, I know his helpfile skills leave mine in the dust and he would actually put the help in helpfile as opposed to myself

If it helps, the patch I've been using to test has changed slightly since the original pr so I've pasted it below

No pressure to make one, it's really only if you want to but it would be very appreciated (:

<pre><code>
1887.3oc2aktaaiCD92o.8cfvnKP5ttZ48QdU1tnP1lIQsVGPVtMYK19rujT
RI9PVlJkRtaBBhAEoLm4iy7MCGx7829lqlsH+A8lYfa.+E3pq9t4IW4dl8IW
09fqlkF+vx0wabCbVbUU41jYya5y1z87pjubedV9Wepmrso4aqVqqbuGp8wE
wkwo5Jc4mzYwKVqscBa6r9EpdrPWKTylA9619RV4lm7Ee9CzYO+sUs79jr69
TodYU86nTjHFRxvXFDpTHIFMGfgjHIhJQbECIL+BUyAHFLBZ5CGAedZLhcR1
SRs6g+6aei8SyGy8FlLeMKzkm.LvuPvXNX1h3r61ATLxZR51z8PXibz9LRGf
GB2G5gwbGlHbPyHBPKySS0YUOIKU5GbhvrqYuOBr7didpAqhqh2nq.k50wUI
4YatOo3o2XcRldY91rp8fz8QZXWVO8B.Ti4CTPgJjxXjfgLNwf.bpEHZLXHz
IEODF7nTGuBbaYdJ3yaxylM.kEMXeEjQ+r9JXj.CQXBVgd1WANkpNcN3ZkQ6
uyXBXzyhsUfz3hBiNrYHP.rOHPJkGCABZGrEJYj8YTNiqDXojglZ73ZhAL9V
YRkFTkOTKAhpWKAiS9wVBHTDkfjDgBwIBJFJtPVBXikfzn72lT4HEFhdK5Su
EXpaUkgoBHSJ4RZ2K+RTjzBPTnjY.KLcxW8MS30bCFbco982oyzkwUOwPND+
ARugOwbbDFgUbJCRULARob.hk9SQin1evJov7WHRLRXPpdyl36zGiANi+ykh
gmYQPPC2gnKpQB8LAIwgGGrwCBDLf6mWfI8EFHiFLjo+lQROFEzOTTBd2sHv
u27WDDFJToWiCJDEAkjm+AQspNKRXxTv3YPoDIUv40TniUJTmz5nnLYGBjeR
fnW5StfGg49jrMAO4lGUfEfEmKE7ifAWF1Gmo8tPBq2DJ3vN7XvpW.j7xMMV
rsp5jYG7B2KVO.Bcv.RWgXw0Doz++iG8tCCITbwwiS4xb65sIqhZRp3GfhR8
pA6.YcdNAt.GNWBi0Qp3P4kfTUuJYYEHIqOXYnzqH0fcc5jdUxlb50Zakh70
OlkmlDutTeWoA7xK+AXgNNaS.MbHC1epa6lyhR+hGLF0eJqTgmVKiXv3II0c
Yev.xUAKBi2VqloVSaYJBIMgXvd.DHuCO.pZ5wC6l1MrlFUMPnAa3zAc4EvP
SNm4FGavs.yF7Ke7C+VxMfe615eCkgRuTkDjHRvrrizHCCIAoLtIFHhPXVai
y4qPCNdX8ShJhW9EWgLtArLe8laFLTX+VRxyhKe7DfB0GPgfcE1BgnStYgCF
tqLeaQ30clWFDRmRy4Se7w04wq.OqEf2gmCBZLStWK90gKPWfpXL9bBvd2OB
mzvITa9+pkEPf7.EdcRBHf9X.75kCfR7Yo+xQAbv1tOclRujMOgf9P.RgtLC
QvQ7Lw8S6SxBnxq7XguQ0UjIWyOn.1+AnLNaUd50l0L3G+na038f+z9ALJPt
B8tQBBSEQQ7m+QZKgsDGwMYLigJNkiwTkz.SDxqjhWx6+5P3LMDOUItItdbs
FHgpX98tuYHa+Eel838pO+BFdx8M19OIlvvfScfND+Jfu4SaXjtIF8XkGAod
cQXNLlP8vb2QkCufUtIy1wAnxl7skKaUhlpb.1Y9Vo2Tkj4tGL6LH09CJMYU
QtQialTLKxFYS4hrIHlM80z542Hubkt7LKggUjsAj7UlIXQuBML.BsMwvyK0
1bHriJDSHzmIrUrBwD50xhUAQAY93dMeAS8X9LcjdM5jlzffGU0HJtqKBhei
LHZFkL.NfPLe+7HoIYERW.jiA8Dfm6NjD.wmf94keks33Gc9sxNN+VQm2dNu
GoUs805BNJVWdYbYK8rkVXPNNpNJ2p83I7YjAQ0vp.XIvkcXIv5X8Uw6nlyC
XjuJstvdE0ACCF4E1q3.GJVGZOau4ydA6bIKRvO+HCmtgNquZ+YCJkDO0MeQ
gvnadE0oc0MDSH1mILfyGx67uviTh4CPDHiTZ1X+1a.deC8mDA7zJB3SmfCh
FoNZexTR8+g.7H1QWLPaei05p+ZDYnZDp9R+Jca.y0bjrLPAHXcmLVW9D3Q7
gri3PLgdEDDQC2D5E6cKNDhIzuzVY8ZtzVWArjraYFbMCiTh8dO9mN4ZLotV
HvZxEdsqnq4Hwm3mX6VyOuXibW2JJFWK11liECRH13G6WxR.TSdedKo9KpVS
s+ZL22sUXDRexNUbt59QaC3XOFR3QMvhcKioq0X4EreB.mPc5Wa3JqOf0m0x
uX9PczAqX66RpAGlTX2p.F4pDNEA6H0FaeikOsWpfWKB0VQclliquNz.xT4U
bl.U63UH4LIivgBNixDXaf.345CKT6EeSbgWuPDuL4Hh8jZayQZKA9I0vgXm
0PPcJKKZ2GUTbQwW0kaZl3ZAdVZ7mycujbdc6jr51r51k5ulz9JMOJtb48IU
5kUaKc54rG3MW+sYo4FQHaaRKIiExrSs6v3xhS0aJhqAG2g1812XFv+AFlUu
S
</code></pre>

tremblap:
help and examples are definitely on @MattS6464 todo (with yours truly) indeed.

lewardo:
@weefuzzy I've been reading into spline regression and wanted to know your thoughts on the interface
1. would the `polynomialregressor` object be suitable as an interface for the algorithm, or would it be preferable to implement the spline stuff in another object, programatically they are _incredibly_ similar, the spline regression is literally a different design matrix and that's about it
1. would natural splines be something I should look into, they are very similar except there are the additional edge case constraints to stop overfitting there
Let me know your thoughts for the UI, I'd be glad to implement either a new object or add the algo to the existing object, although in the latter case it might need a rename

weefuzzy:
My inkling is that `knots` could end up providing some complex interface decisions (like making it easy to have equidistant knots, vs say using k quantiles w.r.t the independent variable). As such, I'd opt for separate user facing objects with the option to do some elegant code reuse behind the scenes, if it seems to make sense. As such, we can defer thinking too hard about it until this simpler object is more done, and then refactor as desired once a putative b-spline regressor becomes a thing. I certainly wouldn't want to make this object more mysterious-seeming by adding `knots`, nor have to think through forms of magic to switch modes.

tremblap:
Hello!

So I get an instacrash with the patch above if I do something naughty (the numberbox to 1 to declare 1 dim, yet having 3 dims generated) then fix then predict (= crash) - otherwise it seems to work fine. Also, we spoke about the 1 to many approach - @weefuzzy thought it was magic but did that grow on him? otherwise this is great thanks

AlexHarker:
@lewardo - is there anything specific you need from me in terms of review? It looks like @tremblap is on testing, so it probably doesn't make sense to double too much, so please direct my attention as needed.

AlexHarker:
It's unclear to me why Owen would have cried - I see no obvious issues with code formatting.

weefuzzy:
Yeah, not unduly tearful (though clang-formatting is always appreciated). I didn't warm to the magic, no; much rather we just made it easier to duplicate / 'broadcast' a dataset column instead

lewardo:
@weefuzzy while talking to @AlexHarker we came upon the subject of allocators, and hence the `ScopedEigenMap` vs raw Eigen object debate for controlled memory allocation - do you think this object requires any custom allocation considerations ? At the moment each object has a raw Eigen matrix for the coefficients, design matrix and filter matrix, would a scoped map be more appropriate ?

weefuzzy:
The simple (or simple-sounding) guideline is: does it ever need to allocate in the audio thread? Eigen's built-in types are hardcoded to use `malloc`, so the ScopedEigenMap stuff is there to let us step around that. 

However, where you have matrix multiplications, it's not always obvious where allocation is happening because Eigen's default behaviour is to create temporaries behind the scenes on the assumption that there is aliasing. If you *know* there isn't aliasing, e.g. `A = BC` then you can signal that with `A.noalias() = B*C`. For more complex expressions it's not always easy to tell though: Eigen has some utility macros you can use to throw asserts when it tries to allocate, which are useful for working out if you need to, say, make your own class-scope temporaries or whatnot.  

(Eventually I'd be inclined to move towards using the ScopedMaps everywhere, so that we can use things like memory arenas for better in-algorithm locality where it makes sense, but that's still a pipedream)

lewardo:
I'll take some time to refactor this to only use `ScopedEigenMap`s just for paranoid safety then, if it's the goal in the long run anyway

Very interesting read about the eigen aliasing, not something I knew about before...

lewardo:
@tremblap's bug has now been fixed, although the fix will be overwritten when I port the eigen stuffs over to the custom allocator - for now however it seems to be a clean object that works, it could be drip-fed to testers for ui testing and the memory things could be done in the bg while the front-end is road-tested

lewardo:
Thank you @weefuzzy for the message, it really clarified my understanding of the ScopedEigenMap/FluidTensor memory allocation paradigm, I am aware that this code now needs refactoring for best memory practice, especially during the matrix multiplications - it allocates correctly but copies the data when it needn't.

I'll finish this at some point, in the meantime the eureka is really helping me with the LSTM object I'm working on (foreshadowing intended), so thank you for the clarification !!

PR #248, opened 2023-08-22
Add a mechanism for communicating streaming size behaviour at compile…

AlexHarker:
… time

This is part of a fix for the max (and other) wrappers that is needed so that wrappers can be aware of how IO sizes will be managed at compile time (and consequently without the need to instantiate a client - as in the main() of the max wrapper or the equivalent class definition section of the pd wrapper).

As this involves hierarchical inheritance this change should not affect the current behaviour of the wrappers in any way until modifications are made to them also.

### DIFF
--- a/include/clients/common/AudioClient.hpp
+++ b/include/clients/common/AudioClient.hpp
@@ -34,12 +34,17 @@ constexpr bool isAudio = isAudioIn<T> || isAudioOut<T>;
 struct Control
 {};
 struct ControlIn : Control 
-{}; 
+{};
 struct ControlOut : Control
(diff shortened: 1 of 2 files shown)


AlexHarker:
@weefuzzy - I've included you here because there are some things to review (or tell us that you don't wish to review) in the max wrapper change coming up and that involves being aware of this change.

weefuzzy (approved):
Sure, fine. One fine day I plan to come up with something more friendly to client authors than tag  inheritance for this stuff, but this change doesn't make that any more or less remote 😄

PR #250, opened 2023-08-24
dynamic time warping algorithms

lewardo:
@weefuzzy @tremblap @AlexHarker 
Beginning of the implementation of the DTW and FastDTW algorithms now that #247 allows time-based manipulation, the pr is mostly for discussion about interface until something functional is built (branch `data-series` merged from lewardo/flucoma-core@54d48c6, hence the slew of commits)

Interface is still up for discussion, since it is merely a complicated but funkier version of a distance between two tensors (take that pythagoras). Algorithms such as `KMeans` would benefit very much from this algorithm, hence `DTW` and `FastDTW` being functors is an interesting paradigm to explore, so that there is not as much refactoring for distance-using objects.

As a proof of concept it will be the distance algorithm used in the `kNearest` interface in `dataSeries` (as provided by `dataSet` with squared euclidian distance)


lewardo:
I have decided to put the implementation of `FastDTW` both on another branch/pr and on the backburner for now, I will however still be continuing the development here of the Sakoe-Chiba and Ikatura constraining windows for the vanilla DTW algorithm

lewardo:
I have now implemented both the Ikatura parallelogram and Sakoe-Chiba band constraints for the vanilla `DTW` object, with the various UI parameters controlling the gradient and radius of the respective constraint. 

I believe that it would be ready to be tested as a proof of concept, to see the reception and whether it would be worthwhile implemented the optimised `FastDTW`

Current messages the `fluid.dtw` object understands are
- `cost <dataseries> <id1> <id2>` which calculates the cost of two series in a `dataSeries` object (by id) and returns a float (the distance)
- `bufCost <buf1> <buf2>` which calculates the cost of two buffers and returns a float (the distance)

other additions:
- `dataSeries` `knearest <buf> <k>` message which, akin to the eponymous equivalent in the `dataSet` object, returns the `k` nearest ids to the buffer passed
- `dataSeries `knearestDist <buf> <k>`, same as above, but returns the k distances equivalent to the k ids above (see the equivalent #252)

both of these are currently constrained using the Sakoe-Chiba band with radius of half the minimum series length - so the width of the band encompasses approximately one third (very roughly) of the points, excluding the diagonal corners (warping the beginning to the end/vice verse), this default was entirely arbitrary and could use some refining although it seems to be sane under testing

Let me know if you have any suggestions or requests !

tremblap:
thanks - interface question incoming!

PR #254, opened 2023-09-04
framework for recurrect neural networks

lewardo:
@tremblap @weefuzzy @AlexHarker 

Now that the `dataseries` object is functional, fun algorithms such as the various flavours of recurrent neural networks can be tinkered with!

The recurrent structure is written in a way to facilitate writing new recurrent networks (e.g. vanilla RNNs, GRUs, adaptive computation LSTMs etc.), currently only an LSTM is being implemented using the framework, but using the `Recur` wrapper (as inspired by the [Flux.jl](https://fluxml.ai/Flux.jl/stable/models/recurrence/) ML library in Julia) writing new algorithms is much easier - I will also write up a guide to writing new recurrent algos as part of this pr once it is fully presentable.

The client is not yet implemented, however there algorithm does compile and regress onto time series successfully - this pr is mostly for now to have a place to discuss potential UI challenges with the client and any other back and forth as needed!


lewardo:
@weefuzzy
Sorry for the poke, but I wanted to ask if you'd ever had any issues with the default `FluidTensor` destructor, I'm getting inconsistent quiet crashes that I've traced back to my destroying of an  `rt::vector` of `FluidTensor`s (technically of structs that inherit from `FluidTensor`, but the same bug also happens with a wrapper) - I've been banging my head against the wall with this and am rather lost - if you don't recognise the issue or have any hunches no worries, I just wanted to know if it was something that rang a bell.

weefuzzy:
> hat I've traced back to my destroying of an `rt::vector` of `FluidTensor`s (technically of structs that inherit from `FluidTensor`, but the same bug also happens with a wrapper)

My hunch is that this is some kind of mixed allocator badness, maybe even combined with some sort of threading nastiness (hence inconsistency). My only advice would be to try and strip it back to the simplest thing you can. E.g. does it still barf with a std::vector container rather than rt::vector? Are you in a position to run with asan? (if this is in Max, then probably not)

lewardo:
I am gladly not in max, Im in a cpp example I wrote (not in the pr) for the algo in core - I've enabled ASan and it is barfing with more detail now - I'll try and make some sense out of it. I'll also strip it down as you said, I'll try and make the smallest bugging example.
One thing that is interesting is that it always barfs if the vector is a local variable, i.e. it gets destroyed at the end of the function call, it is only sporadic when I `clear` a member vector - so perhaps the optimisations in clear cause the inconsistency since I'm running the example in a single thread.

lewardo:
Lots of swearing and head-banging later the issue was the simplest ever - the `FluidTensor` inheritor had a matrix eigen map and an array eigen map, but I didn't take into accout reallocations of the `FluidTensor` pointer so constructed the maps on initialisation, so the maps _sometimes_ would point to a stale pointer if the tensor had been reallocated. Simplest solution is to construct a eigen map when required so it's guaranteed to point to the correct place.

weefuzzy:
> Lots of swearing and head-banging later the issue was the simplest ever 

A happy tale for the ages. FWIW, the inheritance sounds like overkill if you just want a quick way to get an Eigen map over a FluidTensor. I'd like to think that `toEigen` would Just Work as is, but if it doesn't, we should just make some overloads that do...

lewardo:
Now that the object is only really adding `.array()` and `.matrix()` on a `FluidTensor`, I agree that converting it to `asEigen` calls would be wiser - it wasn't an issue with overloads, but rather to do with the maps being preconstructed members in the object rather than dynamically constructed (which turned out to be the whole issue)

lewardo (commented):
great catch on the part of @tremblap,, the first prediction was being discarded and the prediction was skipping points ! 
depending on the input size there can still be issues with output prediction, but if the input length is appropriate given the training lengths then the output is consistent. 
We tested a linear time series and it successfully predicted the correct points so there is definitely consistence convergence (I believe @tremblap still has the patch if anyone would like it)

PR #262, opened 2024-02-07
fixing fluid.normalize json reloading issue

tremblap:
we forgot to compute the mRange in the json reloading... so now it is fixed :)


tremblap:
and obviously, this opens a can of worm. @weefuzzy until we find a way (if ever) to update the @attributes in the client from the json, then I reckon no-change-of-range-post-fitting is the only way forward (as it was for the SC .kr code)

if you are not offended by this, I'll merge this for now to be consistent.

PR #263, opened 2024-02-10
fixes #256 - checking input to env with filters

### LINKED ISSUE #256: `fluid.ampfeature~` can break when it receives a `nan`
As outlined in [this thread](https://discourse.flucoma.org/t/fluid-ampfeature-stops-working-sometimes/) on the discourse when `fluid.ampfeature~` receives a `nan` its own output breaks.

The envelope follower should get pulled in that direction but not ultimately fail.

The object stops working until audio is restarted.

Mac

### Operating system version

macOS 13.4.1

### FluCoMa Version

1.0.6

tremblap:
check the input to env and envgate and stores the last valid value to replace non-finite values with

### DIFF
--- a/include/algorithms/public/Envelope.hpp
+++ b/include/algorithms/public/Envelope.hpp
@@ -71,7 +71,7 @@ private:
 
   double mHiPassFreq{0};
   bool   mInitialized{false};
-  double mPrevValid = 0;
+  double mPrevValid{0};
 
   ButterworthHPFilter mHiPass1;
   ButterworthHPFilter mHiPass2;
--- a/include/algorithms/public/Envelope.hpp
+++ b/include/algorithms/public/Envelope.hpp
@@ -33,10 +33,9 @@ public:
     mInitialized = true;
   }
 
-  double processSample(const double in, 
-                       double floor, index fastRampUpTime, index slowRampUpTime,
-                       index fastRampDownTime, index slowRampDownTime,
-                       double hiPassFreq)
+  double processSample(const double in, double floor, index fastRampUpTime,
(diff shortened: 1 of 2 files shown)


AlexHarker:
See my last comment on the issue. I think there's a better solution.

weefuzzy (changes requested):
Logically looks fine. I have only stylistic consistency nitpicks, at least one of which should have been caught by clang-format: 

* We've been using newline before braces everywhere, IIRC. clang-format should be doing this. 
* Prefer brace-initialization  of the member variable if that's how all the others are done
* Maybe this has just passed the threshold of fiddly-ness where the code duplication between `Envelope.hpp` and `EnvelopeGate.hpp` is tiresome. Could think about pulling out the whole `test if finite and filter-twice-if-needed` saga into its own thing

tremblap:
I have satisfied @weefuzzy first 2 demands. clang-format revealed other discrepancies which I've pushed as we may as well clean it here. As much as I agree with the spirit of the 3rd request, I'd do a proper refactor of these 2 things to make them more efficient in block process (they are doing sample per sample now which is suboptimal)

weefuzzy:
> (they are doing sample per sample now which is suboptimal)

Don't think there's a way around that for these algorithms because they need per-sample feedback. 

But, in any case, is was the formatting consistency that was the thing here. I'd be inclined to go with this unless there's a clear functional reason to play with the other solution ahead of just fixing this bug. Feels to me like a we-can-revisit-if-it's-a-problem kind of thing.

weefuzzy (approved):

tremblap:
ok thanks both. Let's put that fix in and consider other options as future improvements

PR #264, opened 2024-02-27
fix #259 init debounce counter

### LINKED ISSUE #259: `fluid.bufampslice~` not working correctly (when using `@numframes`)
Porting this over here from [the thread I made on the Discourse](https://discourse.flucoma.org/t/fluid-bufampslice-not-working-correctly-when-using-numframes/).

So I’m building [a cascading onset detection thing](https://discourse.flucoma.org/t/chaining-onset-detectors-novelty-ampslice/503/5) using `fluid.bufonsetslice~` followed by `fluid.bufampslice~` and when testing things in context, I’m noticing that `fluid.bufampslice~` is rarely returning valid values when I use the `@numframes` attribute.

In this example if I analyze the whole buffer (`jongly.aif`) I get a bunch of onsets detected (notably not including the first onset), but the moment I add `@numframes 1024` I don’t get any onsets at all anymore. Only when I change `@numframes` to a huge number (`8192`) do I start getting onsets again.

It seems like this is broken to me?

The only thing I can think of is there’s some window being applied when using `@numframes`, but even with that, I would think that adding a `@startframe` offset would compensate for that, but still nothing.

b167370d3686ff8fa697be85da0b7dd3cf7db232

```
1584.3oc0Y0raaiCD9rySAWgdzUf+JRsW1dbeGVTTHaSYqTYICI53jTz9ruC
IkxOM1wzQRF6lCgfzTb32Lbl4aH+wMyhVTeutMB8mn+AMa1OtY1L2P1Al00e
Vz1r6WVl05lVT09sKzMQy8+DzqnpTab+FoavhUtYVu31OyT8ybWVS1VsQ27M
cU1hRscJ3mWk58l9kg1MpeHyC6z98WTzbTzhrp0Qnu9zhZVtonZ82ZzKM9YI
37X7bDUxrMBrqCMFi9p8S94M2X+27.A6x5sa0UldLXz26jRDOFYZd.spHOW2
.S.AHH2hu1XTcU4CnC0MeuEcnvrAstXcVkoXIzqZU8g1+pe0JKpzKq2W4VRd
OhZzsvBlYJpq91qlAqaF40Ul7rk5WpwWreQmJkDpgII5EqWawitulZUWGylf
ewjq.b5Vm+VWdmFPVVzoMGLpy.fkwBngSr8RwVywq15sEqbqIaDMRrXTUswt
ePG1nq7FlrprxGdzNFbfzfpyQlMZzh8VC4QMLzAYXNqYPbcLCBVp0.HncNGt
FdxH6VPiQs0nsYUO.J6VXO+GQCQ2vuN5FBVHhkDIOUjxUXYBkqliHJVLmkR3
ILEmR3JlzNXpS+8QUbP.h1r052n3HDElGezfpzinXXQGUCPNYfy2KfImKcQJ
YhQHf4IAnUKFcVWIZnZ.xXpAnJmFf.NGfOBmNIZ.EIkFrAFOEvKgXgm2LO5v
iiSSBEdzzI.dTLaBgGEyUACO0T.O9XX8pzGfc3aPmAs.UDJkNpL5xns4nrAT
2Jr93mGoLhOdjyYb7CDgo7fMjISggTplvyoGxtSmW2rEcac0ZfV0bTaYwRcK
5Sz9g5+lMEqVoqNYB4ioO3io9PIowoLdZJjTMg.LSRS.EBywLgjJGj54118A
edlkd4knLlHtqPE648DAFqHXHUoTf4fZP.JfDYZJlB8YBkP7z5mWTpehlSd4
9hUw8F+eEeaazHRqiDir9vHScG8Ys+rzCCgcGkdcX2wA+MaTElu.Dwf3ucJM
jI66ZTdQSqwS78cqj3bZFxIxefeG58Xk2ewEaQH6P73BxVishoLTYgA1SnEE
FjNqorP2ftcOfb33wBMpMKWOL3mbwvm6okyHtbjDJcP3ewdiotJzvGD9HF93
72tg2P2QVuqZN93xFvGLAJIlVVzZBMYI4R4r5V7.JLAJK6sEhImFR5ti3tKx
A8Ixbjm4RX3WMA0jwDNyLgvlBVeeFQEAyY+RoBcNxd8HzS1ikLL.dQ2Io3Ze
kjcXcbtRxSXLerzyzCQBlcazkC0ffomBmXRpTweQc+pmKaGOj3rh7fiWegvN
urNyDjYVk3iM6BMQXCiA6kkDhxUWyrP8NucbplzjPYa24NX+KzWZq22rT+js
+K4YsF.r61uCwdt2p5CUHlBFosr9P2uSoXwyC3lheHfN4lFc6l5xUPbVned9
yC.+Nn2chGHau1rA8xxG+Pranb9D42ontauRg8YNF662QAmqQxDfMMTVhPwh
AG8jDIzj.0x.MRYBN1xASZa.hJvuA4rg5ZfMEmBSQPRRsMbnLmXTBX.fJKRR
RsKlDS.t+HIPWGFTJgRchQJgDHriRYTNLkTNm.SgfwV9dHacRPdCnUQs6MHK
o6aIjjD6zIThPY2sTtsl6vsZGKVIabSu6yDPfsU5uyrQzUuh3+Bj5nL9Dwpi
vwiNqtADJ4o2uy4g++zPKO+XRibnEN6XGTE7AcPMyXZtfao4imeyJndZhdib
zGyqM0CXHFlSYHR69ycSMGmUma8blye6Q1cquc7Wqg7mK61M8OfN54WKckt0
TT4d4jWLIfbNhbBiPvBJ.43NggvCTRpqljdZUdWYYqbc3hhGDrrY+GAYwCRE
95YU2rR67CHWCg+5I816Z9ocCdP6lda26tYruWp0+XbUD12f775A4HXuY3qk
jruL2URRpqljjAcdUMVhhbtSiigfRtZZuP71Yomxae5ctImHJ2vhqDh9cLNw
HBL69fMiAHmwIiaPYGljDSgIZGgJx6edwyZKa2t6zMsceuSt.C1aqcSWM20s
nx2U351nuqne99QxZ.VjFfB49F+SVcehmjez1ZPvU.AXOvADChzwN197Vs65
dNMGI5a94M+aWnL7.
```

tremblap:
Fixes #259 by setting the init of the debounce counter in the 3 slicers that had it to 0 instead of 1. when 1, the condition when the (infinite state of 0 input) to (any signal) which should correctly trigger an attack at frame 0 was ignored, which is a conceptual error compare to our RT one.

Conceptually, we always assumed silence padding as it is what happens in the CCE in RT to these inputs. It is the less worse option. that counter was starting at 1 and dismissing signal at input.

see the bug report to have a comparison patch. or just approve :)

thanks

### DIFF
--- a/include/algorithms/public/EnvelopeSegmentation.hpp
+++ b/include/algorithms/public/EnvelopeSegmentation.hpp
@@ -28,7 +28,7 @@ public:
   void init(double floor, double hiPassFreq)
   {
     mEnvelope.init(floor,hiPassFreq);
-    mDebounceCount = 1;
+    mDebounceCount = 0;
     mPrevValue = 0;
     mState = false;
   }
@@ -63,7 +63,7 @@ public:
 
 private:
   Envelope mEnvelope;
-  index  mDebounceCount{1};
+  index  mDebounceCount{0};
   double mPrevValue{0};
   bool   mState{false};
 };
--- a/include/algorithms/public/NoveltySegmentation.hpp
+++ b/include/algorithms/public/NoveltySegmentation.hpp
@@ -36,7 +36,7 @@ public:
       Allocator& alloc = FluidDefaultAllocator())
   {
     mNovelty.init(kernelSize, filterSize, nDims, alloc);
-    mDebounceCount = 1;
+    mDebounceCount = 0;
     mPeakBuffer.setZero();
   }
 
@@ -64,7 +64,7 @@ public:
 private:
   NoveltyFeature          mNovelty;
   ScopedEigenMap<ArrayXd> mPeakBuffer;
-  index                   mDebounceCount{1};
+  index                   mDebounceCount{0};
 };
 } // namespace algorithm
 } // namespace fluid
--- a/include/algorithms/public/OnsetSegmentation.hpp
+++ b/include/algorithms/public/OnsetSegmentation.hpp
@@ -38,7 +38,7 @@ public:
   void init(index windowSize, index fftSize, index filterSize)
   {
     mODF.init(windowSize, fftSize, filterSize);
-    mDebounceCount = 1;
+    mDebounceCount = 0;
   }
 
   /// input window isn't necessarily a single framre because it should encompass
@@ -66,7 +66,7 @@ public:
   }
 
 private:
-  index                   mDebounceCount{1};
+  index                   mDebounceCount{0};
   OnsetDetectionFunctions mODF;
   double                  mPrevFuncVal{0.0};
 };


tremblap:
the RT worked only because one would need to start the DAC and the play at the same time to 'miss' that attack. So I think we just never caught it. With this fix, it works both in RT and NRT.

weefuzzy:
So now the slicers will always emit an event at `t==0` if the first sample they see is greater than the threshold?

tremblap:
yes (not the sample, but the value of the algo - in amp it is a diff between properly initialised env, I've checked) - this is what we expect in RT (from silence) so not having that (from assumed padded silence) in NRT was problematic (for instance in Rod's case)

weefuzzy:
Well, if it still passes all your detailed supercollider tests, have at it

AlexHarker:
As far as I can see the code as it was prevented slices at t==0 and it's unclear if that was a design, or an accident. However, I don't see a clear rationale for disallowing that in either RT or NRT case (and in the latter is more likely problematic). I think the change here is probably better default behaviour.

g-roma:
Clearly this was by design, explicitly to avoid detecting an onset at the first sample. I think the rationale was linked to the padding which happens outside the object. I am not sure whether this may have changed. The design was based on a number of test cases, sometimes with potentially conflicting requirements, so it was "the less worse option" at the time, but if this passes all of these now then I guess it should be fine.

tremblap:
it passes all the unit tests so I'd merge in

PR #265, opened 2024-03-01
Update flucoma-core for hisstools header-only

AlexHarker:
This is a PR that updates code and CMake files to deal with HISSTools_Library in header only format, which will be the future of the library.

The changes in the repo have been tested for compilation, but it would be advisable to run some audio tests for sanity in the SC repo and also do some audio testing. In terms of impact on the flucoma ecosystem this pulls in the header-only version of the FFT, which is edited from the previous version - there is no specific reason to expect any issue, but any obvious error is likely to turn up in basic audio testing pretty easily.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -56,7 +56,7 @@ FetchContent_Declare(
   HISSTools
   GIT_REPOSITORY https://github.com/AlexHarker/HISSTools_Library
   GIT_PROGRESS TRUE
-  GIT_TAG 447421d
+  GIT_TAG f3292ad 
 )
 
 FetchContent_Declare(
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -56,7 +56,7 @@ FetchContent_Declare(
   HISSTools
   GIT_REPOSITORY https://github.com/AlexHarker/HISSTools_Library
   GIT_PROGRESS TRUE
-  GIT_TAG ce4c7f6
+  GIT_TAG e2d8083
 )
 
 FetchContent_Declare(
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -56,7 +56,7 @@ FetchContent_Declare(
   HISSTools
   GIT_REPOSITORY https://github.com/AlexHarker/HISSTools_Library
   GIT_PROGRESS TRUE
-  GIT_TAG e5091cf
+  GIT_TAG ce4c7f6
 )
 
 FetchContent_Declare(
@@ -168,21 +168,6 @@ if(NOT fmt_POPULATED)
 endif()
 add_subdirectory(${fmt_SOURCE_DIR} ${fmt_BINARY_DIR} EXCLUDE_FROM_ALL)
 
-# HISSTools FFT target
-add_library(
-  HISSTools_FFT STATIC "${hisstools_SOURCE_DIR}/HISSTools_FFT/HISSTools_FFT.cpp"
-)
-
-target_link_libraries(
-  HISSTools_FFT PRIVATE ${ACCELERATE}
-)
-
-target_include_directories(HISSTools_FFT PUBLIC "${hisstools_SOURCE_DIR}")
-
-set_target_properties(HISSTools_FFT PROPERTIES
-    POSITION_INDEPENDENT_CODE ON
-)
-
 # Brute force staic runtime on windwos
 if(MSVC)
   foreach(flag_var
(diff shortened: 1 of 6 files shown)


AlexHarker:
It would be wise also to clang-format any relevant files.

tremblap (changes requested):
The SC repo doesn't compile - loads of:
> clang: error: linker command failed with exit code 1 (use -v to see invocation)

so there must be a few changes needed in wrappers. please advise.

sclang_format was done then undone for most of it.

tremblap:
(and thanks for doing this for @fearn-e I reckon :) )

tremblap:
I am far from a CMake specialist, but commenting lines 136-138 in CMakeLists.txt in the SC wrapper seems not to sort it. Am I supposed to find other problems? Should I do similar unblinking (if that is what I do there) in the other 3 repos?

tremblap:
I'm also commenting line 45 of MakePluginSource.cmake - it seems to behave better. Can you please have a check in the SC repos (to start with) to see if that is the only problem @AlexHarker - then we can troubleshoot the other 3 repos (max, pd, cli) and do PRs in all 4 wrappers and then we have a potential merger.

tremblap:
(it compiled at last, running the SC UTs in MacOS now)

tremblap:
ok this is working perfectly now. I'll wait for approval of code on the other repos - I'm pushing them now

tremblap:
it seems only SC needed some manual linking. @weefuzzy am I wrong? The other 3 CCE wrappers compile... so far :)

AlexHarker:
You've only given the most outer error (linker fail) but it's the detail that I'd need - what was/is the symbol that fails to link?

The most obvious thing to fail would be linking to accelerate, but as I'm not a cmake expert I don't know whether that's happened - there aren't links to the lines you cite above (which would have been helpful) but I'll go and look and see if any of this makes sense of why it would now link...

AlexHarker:
OK - thanks - the PR for sc clarifies this - those would be required changes (removing references to HISSTools_FFT in cmake in the dependent repos). I didn't experience any required changes in the max cmake, but we would need to check in all 4? repos for each CCE plus command line

tremblap:
already there :) Max compiled, Pd compiled, cli is failing with an easy one :
#include <AudioFile/IAudioFile.h>
I'll change that and make a PR

weefuzzy:
Off the top of my head I can't think of any reason flucoma-sc would need special treatment for this. As @AlexHarker says, we'd need to see the error diagnostics to to be more helpful. If you're using `iTerm` and have output capture set up, being able jump to bits where it says `error` is refreshingly easy. 

FWIW, I'm surprised that flucoma-cli is building (?) without further changes, given that it uses all the hisstools audio file stuff.

AlexHarker:
SC needed -fPIC - PA has dealt with the changes there.

The CLI *shold break* (pd and max are fine) but is it pulling in HISSTools separately to core (and therefore on its own version). That repo needs code updates as @weefuzzy rightly points out - I did quick searches of the repos on GitHub to confirm all the above.

AlexHarker:
Oh good to know that the CLI is failing (missed that above) - @tremblap  - you can try to follow similar changes to the PR here, but if you struggle I may need to do that one, as there are also changes to snake case and method renamings (plus the namespace) that will mess it up. describe.cpp or the single tests template file in this PR and the ones to look at.

tremblap:
Let's fix CLI before we merge. I'm trying to find all the new spiel

tremblap:
describe is good but only has #include <AudioFile/IAudioFile.h>

there is also #include <AudioFile/OAudioFile.h> in the wrapper, plus calls to BaseAudioFile

there are only 6 instances of HISSTools::* so if you could look at it now that would be ace (and allow me to adapt the other 2 examples in core in a PR that will break)

tremblap:
it would be good if CLI wasn't calling its own version of it too - just so we are sync now (it might diverge again in the future but hey, sync is good)

AlexHarker:
No the CLI pulls it's version from the cmake in core, so it is aligned (cmake builds are separate, however). I'm updating the CLI now, but it looks like the out file code has not been tested anywhere so I'd advise some sanity checking (at least through listening) to the output of the CLI.

tremblap:
ok it all works now thanks for this

AlexHarker:
Thanks for providing the testing to improve/sanity check header-only hisstools library!

PR #279, opened 2024-10-09
Tree making overrun in UMAP - fixes #277 and #278

### LINKED ISSUE #277: Instacrash when `fittransform`-ing in `fluid.umap~`
I remember this being an issue a while back but it being fixed (hence the `fluid.umap~: Number of Neighbours is larger than dataset` error message now when trying to `fittransform` something that has fewer dimensions than `@numneighbours`) _but_ there appears to have been an oversight when coding it as it gives a nice instacrash when `fittransform`-ing a dataset with the _same_ amount of dimensions as `@numneighbours`.

Screenshot 2023-07-03 at 11 12 04 AM
[patch.zip](https://github.com/flucoma/flucoma-max/files/11935677/patch.zip)

I imagine this means that a `>` needs to get changed to a `>=` when running the size check.

Also including three crash reports, two of which point to a gnarly UMAP error but the third one is different (no flucoma mention at all), so including it in case it's indicative of something else happening at the same time.

[crash reports.zip](https://github.com/flucoma/flucoma-max/files/11935688/crash.reports.zip)

tremblap:
The tree-making seems to need to remove the front of the matrix - there is a boolean in the algo, but since tree making can be run on a square matrix elsewhere in the codebase, this was creating an overflow in UMAP as the client was not checking for a smaller number of neighbour than the number of points.

@weefuzzy and @g-roma if the rejection of the first item at UMAP.hpp:373 in fluid::algorithm::UMAP::makeGraph is right for the algo (I am not certain) then this fixes everything.

### DIFF
--- a/include/clients/nrt/UMAPClient.hpp
+++ b/include/clients/nrt/UMAPClient.hpp
@@ -79,8 +79,8 @@ public:
     auto src = srcPtr->getDataSet();
     auto dest = destPtr->getDataSet();
     if (src.size() == 0) return Error(EmptyDataSet);
-    if (get<kNumNeighbors>() > src.size())
-      return Error("Number of Neighbours is larger than dataset");
+    if (get<kNumNeighbors>() >= src.size())
+      return Error("Number of Neighbours is greater or equal to the size of the the dataset");
     FluidDataSet<string, double, 1> result;
     try
     {


weefuzzy:
Certainly not familiar enough with the internals to pronounce authoritatively, but support for this change comes from the fact that this is what the reference implementation checks for too: 

https://github.com/lmcinnes/umap/blob/c72ac2f4bc1b107a055a30e9582b37ea881d95ca/umap/umap_.py#L2430

g-roma:
The number of neighbors obviously needs to be smaller than the dataset. I think the reason for discarding the first neighbor is that it is the same as the query point, but IIRC it is used when transforming new data, so the option is there. But in the fitTransform case it is true that the number of neighbours needs to be at most (size -1), to account for the query point when making the graph.

tremblap:
ok then I'll merge, thanks for the explanation

PR #280, opened 2024-10-09
removes the word rows from ds.print and the interface of dsquery.limit

tremblap:
see related PRs in all repos (core, docs, sc, max, pd)


tremblap:
The motivation: every time I read the print out, I get confused. Since they are called points, and cols, then let's unify them.

tedmoore (approved):

jamesb93 (approved):

jamesb93:
Looks good!

You might like this feature of `git`, but it's more of a detail in the scheme of things.

https://blog.sebastian-daschner.com/entries/git-commit-fixup-autosquash

PR #281, opened 2024-10-11
fix max fft size fiasco

### LINKED ISSUE #258: `fluid.bufnmf~` crashes with fft size larger than 65k
### Please tell us what you were doing! You can include code and files by drag and dropping them into the text area.

So was trying to do something with `fluid.bufnmf~` where I analyze using really large fft windows hoping to catch changes in "material" in there, and found a reproducible instacrash.

I tried to do a 3 component NMF decomposition using `@fftsettings 131072 2048 131072` on `jongly.aif` (which is only 124438 samples long) and it instantly crashes.

It should analyze the buffer with these settings, perhaps zero padding the end of the window, or at least return an error saying the file has to be `<` the FFT size.

Instacrash.

Mac

### Operating system version

macOS 13.4.1

### FluCoMa Version

1.0.6

weefuzzy:
What is the road to where paved with? 

This hopefully fixes #258 

`FFT` had been altered some time ago to use a global `setup` object in order to cut down on these expensive things being redundantly created all over the place (quite possibly on the heap? Can't remember...). Instead an assumed hard limit of 2^16 Became A Thing. Unfortunately, nobody (i.e. me) told the rest of the codebase, which happily assumed that anything was possible. Moreover, (a) `FFT::resize()` is assumed by calling code not to fail (!?) and (b) its `assert` was useless because `FFT` was still cheerfully accepting fft sizes > 2^16, even if its `setup` object could never honour that. 

(Seems like now would be a good time to have more unit tests) 

So. I still like the idea of 'normal' FFT usage not spraying lots of these `setup` objects out into the world, but have attempted to accommodate the adventurous by allowing a local `setup` object in a `std::optional` when a chunkier size is requested.    

So far, this is tested only to the extent of seeing if I could run NMF in Max with a FFT of 2**17. I will at least confirm that when fft size is <= 2**16, it continues to use the default path, but more dedicated tyre kicking by @tremblap will be helpful.

### DIFF
--- a/include/algorithms/util/FFT.hpp
+++ b/include/algorithms/util/FFT.hpp
@@ -39,8 +39,8 @@ public:
   FFTSetup(FFTSetup const&) = delete;
   FFTSetup& operator=(FFTSetup const&) = delete;
 
-  FFTSetup(FFTSetup&& other) { *this = std::move(other); };
-  FFTSetup& operator=(FFTSetup&& other)
+  FFTSetup(FFTSetup&& other) noexcept { *this = std::move(other); };
+  FFTSetup& operator=(FFTSetup&& other) noexcept
   {
     using std::swap;
     swap(mMaxSize, other.mMaxSize);
--- a/include/algorithms/util/FFT.hpp
+++ b/include/algorithms/util/FFT.hpp
@@ -63,8 +63,6 @@ class FFT
 public:
   using MapXcd = Eigen::Map<Eigen::ArrayXcd>;
 
- //TODO: why is this here? 
- static void setup() { getFFTSetup(); }
 
   FFT() = delete;
 
--- a/include/algorithms/util/FFT.hpp
+++ b/include/algorithms/util/FFT.hpp
@@ -14,6 +14,7 @@ under the European Union’s Horizon 2020 research and innovation programme
 #include "../../data/FluidMemory.hpp"
 #include <Eigen/Core>
 #include <fft/fft.hpp>
+#include <optional> 
 
 namespace fluid {
 namespace algorithm {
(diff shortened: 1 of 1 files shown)


weefuzzy:
The ubuntu CI is failing because one of the tests won't build. Seems like GCC has decided that in that test I'm trying to copy a move only object, but not at all clear why it's decided that now because this was always the case.

tremblap:
oh ubuntu is bailing at the tests (any idea why?), but I ran the SC unit tests and they all run fine...  and nothing crashes now with the large fft size.

tremblap:
otherwise I should test on ubuntu and windows as this is potentially 'big' right?

tremblap:
I mean, running the UTs on locally compiled things?

weefuzzy:
Yes, whenever you're able!

tremblap:
doing it now as I practice this hard bass line I can't play to save my life :D

tremblap:
ok it refuses to build on Ubuntu 20. I get loads of similar errors to this:

/home/pa/Documents/source/flucoma-core/include/clients/rt/../common/../../algorithms/public/STFT.hpp:27:7: error: use of deleted function ‘fluid::algorithm::FFT& fluid::algorithm::FFT::operator=(fluid::algorithm::FFT&&)’
In file included from /home/pa/Documents/source/flucoma-core/include/clients/rt/../common/../common/FluidContext.hpp:17,

tremblap:
I've pulled the tip of main in SC, and the PR 281 in core, and building in ninja. I can send the full error stack but there are a lot.

weefuzzy:
e-mail it to me if you can. sounds horrible

weefuzzy:
Ok, received, thanks. That does give me a clue, now off to cppreference to see if I can figure out what exactly I've done to make gcc so sad (it's definitely this change, to do with adding a `std::optional` to the FFT class).

tremblap:
in case this helps explaining why MacOS doesn't bail but Ubuntu20 does: on the latter I have gcc version 9.4.0 (Ubuntu 9.4.0-1ubuntu1~20.04.2)

weefuzzy:
Ok, think I figured it out. Can you check a full build again?

tremblap:
It builds in all 3 OSes, I am running the UTs in SC now in all 3

tremblap:
it all works perfect, you can merge

tremblap (approved):

Issue #274, opened 2024-05-19
Bug: instacrash of `fluid.knnclassifier~` when loading empty dictionary


rconstanzo:
### Please tell us what you were doing! You can include code and files by drag and dropping them into the text area.

I found an instacrash when you try and `load` a `dict` into `fluid.knnclassifier~` that was created by a `fluid.classifier~` *prior* to having fed it any data.

When you `dump` or `write` a `fluid.knnclassifier~` without data empty you get the following dictionary contents:
```
{
	"labels" : 	{
		"cols" : 1
	}
,
	"tree" : 	{
		"cols" : 0,
		"data" : null,
		"ids" : [  ],
		"rows" : 0,
		"tree" : null
	}

}
```

Here's a patch to reproduce:
![Screenshot 2024-05-20 at 12 01 56 AM](https://github.com/flucoma/flucoma-core/assets/2747015/1dece0a6-d5e3-4ec8-ada3-8463f8703403)

```
<pre><code>
----------begin_max5_patcher----------
723.3ocwV0zbZCCD8r8uBUclxfr4iPO0i8R+EzoCivdgnTaIOxxDRxj9auRq
rHl.NAHz1CVy30qzaeO8zZ8TbDcoZKTSIeg7CRTzSwQQXHWfn12ink7sYE7Z
LMZlprDjF5.+2LvVCFOYHISyquk7c91OE9ZgPBYpFIlRRavUJooV7H3hwlNb
TmvRdIFl9MnXCXDY7vJIaJExBvfEAqMXE2jcqPtdgFxLdRjlZWPxTlab1Mtw
T6H4murLpFSXc5B8JdFzcsE4XgnVd2mmQcgdNN1ML3CpSrgjMB3dRtHyPxr.
aSp9+mdwtAErjYtQFK85nXoWhhsrwXTxyk.SvRlMZLRici8T+gExGx7PE3WE
5RtbMc2j5PkozcXqspsAzK.IeYANwQW.Mkv8108.ewiEDMr9nrO4M19Rlize
xrvFmk9I8R+j9nOcf84XzmcI6jNu8PmK+b2LS7GbYiQ5jjN0StouqerSEO4e
m2iM12g4uk2a7+HuG1KR1TTP9JTtDxIry1G5Uh4I91Hr2wGNoOovUIBkjqe.
cjceNl.kjFJzbtg2R7Vla6lxWBE0gngvQ1Vy9nLe.q70NAiFf9Su0q0ALmjs
[...]

PR #282, opened 2024-10-12
Crashes with JSON generated by empty objects

weefuzzy:
Motivating incident was discovering that with an 'empty' `KNNClassifier` we can generate JSON that will then cause a crash on read (#274)

First commit here 'fixes' that by identifying such JSON as invalid and error-ing out instead (`KNNClassifier` JSON is combination of `KDTree` and `Dataset` but only the top level was being checked. Now we check the sub-objects by delegating to their JSON validation fns) 

Now we need to check the other serializable objects too. 

- [x] DataSet
- [x] DataSetQuery
- [x] LabelSet
- [x] KDTree
- [x] KMeans
- [x] SKMeans
- [x] KNNClassifier
- [x] KNNRegressor
- [x] Normalize
- [x] RobustScale
- [x] Standardize
- [x] PCA
- [x] MDS
- [x] UMAP
- [x] MLPRegressor
- [x] MLPClassifier
- [x] Grid

### DIFF
--- a/include/clients/nrt/KNNClassifierClient.hpp
+++ b/include/clients/nrt/KNNClassifierClient.hpp
@@ -43,7 +43,10 @@ void to_json(nlohmann::json& j, const KNNClassifierData& data)
 bool check_json(const nlohmann::json& j, const KNNClassifierData&)
 {
   return fluid::check_json(j, {"tree", "labels"},
-                           {JSONTypes::OBJECT, JSONTypes::OBJECT});
+                           {JSONTypes::OBJECT, JSONTypes::OBJECT}) &&
+         fluid::algorithm::check_json(j.at("tree"), algorithm::KDTree()) &&
+         fluid::check_json(j.at("labels"),
+                           FluidDataSet<std::string, std::string, 1>());
 }
 
 void from_json(const nlohmann::json& j, KNNClassifierData& data)
--- a/include/clients/nrt/KNNClassifierClient.hpp
+++ b/include/clients/nrt/KNNClassifierClient.hpp
@@ -22,7 +22,7 @@ namespace knnclassifier {
 
 struct KNNClassifierData
 {
-  algorithm::KDTree                         tree{0};
+  algorithm::KDTree                         tree{algorithm::KDTree()};
   FluidDataSet<std::string, std::string, 1> labels{1};
   index size() const { return labels.size(); }
   index dims() const { return tree.dims(); }
--- a/include/clients/nrt/KNNClassifierClient.hpp
+++ b/include/clients/nrt/KNNClassifierClient.hpp
@@ -24,9 +24,9 @@ struct KNNClassifierData
 {
   algorithm::KDTree                         tree{0};
   FluidDataSet<std::string, std::string, 1> labels{1};
-  index                                     size() const { return labels.size(); }
-  index                                     dims() const { return tree.dims(); }
-  void                                      clear()
+  index size() const { return labels.size(); }
+  index dims() const { return tree.dims(); }
+  void  clear()
   {
     labels = FluidDataSet<std::string, std::string, 1>(1);
     tree.clear();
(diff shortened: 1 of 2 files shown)


weefuzzy:
Nothing else crashes. Everything with the exception of `KNNRegressor` complains about invalid JSON.

weefuzzy:
Ok, `KNNClassifier` and `KNNRegressor` both behave differently to everything else due to one of those special C++ subtleties: the shared JSON generating function will just return an empty string if the underlying object isn't initialised (we can discuss if that's what we really want later), however because of the way that the `KDTree` belonging to these two objects is created, they were returning `true` for this check even when it wasn't, like, really true. 

Gory details: `KDTree` has a default constructor that won't initialise the object, and one that takes a `Dataset`, which will. The `KNN*` objects are making the tree by calling `KDTree tree{0}` which C++ then heroically manages to interpret as an implicit cast of `0` to an empty `Dataset` (we can also discuss if we really want that), so then the tree is marked as initialised. We can simply change this to call the default constructor instead and then the `KNN*` objects behave like everything else.

tremblap:
This fixes the crash for KNNclass but not KNNreg. But this code reveals a mess of default save states which maybe we should investigate eventually?

```
// save all empties
["DataSet", "LabelSet", "KDTree", "KMeans", "SKMeans", "KNNClassifier", "KNNRegressor", "Normalize", "Standardize", "RobustScale", "PCA", "MDS", "UMAP", "Grid", "DataSetQuery", "MLPClassifier", "MLPRegressor"].do{|x|"Fluid%(s).write(\"/tmp/%.json\")".format(x,x).postln.interpret};

// open them
["DataSet", "LabelSet", "KDTree", "KMeans", "SKMeans", "KNNClassifier", "KNNRegressor", "Normalize", "Standardize", "RobustScale", "PCA", "MDS", "UMAP", "Grid", "DataSetQuery", "MLPClassifier", "MLPRegressor"].do{|x|"/tmp/%.json".format(x).postln.openOS}

// load all empties
["DataSet", "LabelSet", "KDTree", "KMeans", "SKMeans", "KNNClassifier", "KNNRegressor", "Normalize", "Standardize", "RobustScale", "PCA", "MDS", "UMAP", "Grid", "DataSetQuery", "MLPClassifier", "MLPRegressor"].do{|x|"Fluid%(s).read(\"/tmp/%.json\")".format(x,x).postln.interpret};
```

tremblap:
I think a discussion on what is a valid saving of a valid empty state for each is needed... MLP regressor is quite funny... in the meantime I will modify the check on KNNreg

weefuzzy:
You'll notice that I haven't pushed anything touching `KNNRegressor` yet ;-) 

When you say a mess of states , can you elaborate? AFAICT, with the exception of `KNN*` everything else is emitting an empty JSON string: are you seeing side effects on new objects when you read that back? In Max it seemed to be consistently saying 'invalid blah'

tremblap:
I'm doing the regressor for the understanding (and will also run clang-format)

mess of states: check MLPReg's content of saving a default state... but no UX difference you're right - it just bails.

weefuzzy:
if I do in SC

```supercollider
m = FluidMLPRegressor(s); 

m.write("/Users/owen/Desktop/mlp.json")

n = FluidMLPRegressor(s); 

n.read("/Users/owen/Desktop/mlp.json")
```
Then 

* the resulting json file just contains the word `null` 
* attempting to read it says `invalid blah`

Is that what you see?

tremblap:
pushed the KNN and clang-formatted

> Is that what you see?
yes. But I'm talking about the content of those json files. the mlp is "null" - same in mlpclass, its mlp is "null", and maybe that is right. I'll mull over it when I start correcting the other bug on 'real check' and will obviously ask you for wisdom on what we should produce and check.

in all cases, if you're happy with the KNN I've pushed and the formatting, let's merge this

tremblap:
oh it didn't push

weefuzzy:
Arg. I already had a local commit 😢 

(Also, can we not format entire files as side effects of small changes pls? It's possible to apply formatting just to a selection and it makes the commit easier to read)

weefuzzy:
The tree initialization loophole for `KNN*` now fixed in last commit, so now everything should behave consistently, if not desirably.

weefuzzy:
`write` and `dump` don't behave equivalently: `write` doesn't check that the algo is initialized, it just serializes unconditionally. 

https://github.com/flucoma/flucoma-core/blob/3ceddcd10b26573fb20260e335f7272619e8ddd7/include/clients/nrt/DataClient.hpp#L41-L46

https://github.com/flucoma/flucoma-core/blob/3ceddcd10b26573fb20260e335f7272619e8ddd7/include/clients/nrt/DataClient.hpp#L61-L67

tremblap:
> (Also, can we not format entire files as side effects of small changes pls? It's possible to apply formatting just to a selection and it makes the commit easier to read)

ok sorry

tremblap (approved):

tremblap:
compiling works, and last tests work too

tremblap:
merge at will

PR #284, opened 2024-10-27
TruePeak Interpolator: Ensure enough storage

### LINKED ISSUE #283: Crash: `Interpolator` doesn't allocate enough memory at higher SRs
Crash in SC server reported by user here: https://discourse.flucoma.org/t/96khz-seems-to-hangup-the-server/2438

Can reproduce on my mac. Cause turns out to be that at higher sample rates, the buffer for the `Interpolator` used by `Loudness` is undersized, because I clearly wasn't thinking that day. 

The offending line is 

https://github.com/flucoma/flucoma-core/blob/ef77dbafc2735224505a42d05e78c3140827edb4/include/algorithms/util/TruePeak.hpp#L40

which doesn't take into account that for lower upsampling factors for a given number of taps, the max latency would end up being bigger, viz. (49 + 2 -1) / 2 > (49 + 4 - 1) / 4 🤦 

Upshot is that the class doesn't allocate enough storage at 88/96 and so crashes.

weefuzzy:
fixes #283

Introduce a min upsampling factor to ensure that maximum delay can be anticipated at construction time (would be better, actually, if SR were passed to `TruePeak` at construction time, but one thing at a time).

### DIFF
--- a/include/algorithms/util/TruePeak.hpp
+++ b/include/algorithms/util/TruePeak.hpp
@@ -35,10 +35,12 @@ class Interpolator
   };
 
 public:
-  Interpolator(index maxtaps, index maxfactor, Allocator& alloc)
-      : mMaxTaps{maxtaps}, mMaxFactor{maxfactor},
-        mMaxLatency{(mMaxTaps + mMaxFactor - 1) / mMaxFactor},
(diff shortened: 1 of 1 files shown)


tremblap (approved):
ran the UTs - results as expected

PR #286, opened 2024-10-31
Enhancing K-Means initialization options

weefuzzy:
We're finding that `SKMeans` can perform quite poorly with the random partition method of initialization, insofar as it is prone to get stuck in local minima. 

This can be a problem for k-means more generally (see [1]), depending on the distribution of the data: k-means's hard assignment means that a poor choice of initial centroids is more likely with random partition and unlikely to converge well in the subsequent iterations. 

This PR explores improvements. First by adding an option to sk-means for 'Forgy' initialisation, which just initializes the means by sampling `k` points from the data. This seems to produce improvements for sk-means in some simple cases. However, it's still not the state of the art in initialization procedures, because it depends on how representative those `k` samples are. 

More sophisticated schemes such as k-means++ build up a more accurate approximation of the data distribution and sample from that. k-means++ is expensive however. Fortunately there are now more efficient approximate schemes for this that use MCMC to approximate the distribution with only a single pass over the data. 

This PR will stay WIP for a bit whilst we explore this in a controlled way. This initial commit aims at a minimal set of changes to skmeans to provide the choice of initialisations. I imagine the following steps: 

- [ ] verify that `Forgy` initialization  can improve matters for sk-means 
- [ ] refactor k-means a bit so that it can offer a choice of init schemes too 
- [ ] develop and test MCMC initialisation for both k-means and sk-means 

----

[1] Greg Hamerly and Charles Elkan. 2002. Alternatives to the k-means algorithm that find better clusterings. In Proceedings of the eleventh international conference on Information and knowledge management (CIKM '02). Association for Computing Machinery, New York, NY, USA, 600–607. https://doi.org/10.1145/584792.584890

### DIFF
--- a/include/algorithms/public/SKMeans.hpp
+++ b/include/algorithms/public/SKMeans.hpp
@@ -18,6 +18,7 @@ under the European Union’s Horizon 2020 research and innovation programme
 #include "../../data/TensorTypes.hpp"
 #include <Eigen/Core>
 #include <queue>
+#include <random>
 #include <string>
 
 namespace fluid {
@@ -27,8 +28,16 @@ class SKMeans : public KMeans
 {
 
 public:
+
+  enum Initializer { 
+    // Random partition assigns points to random clusters at init
+    Random_Partition,  
+    //'Forgy' initializes means with k random data points
+    Forgy   
+  }; 
+
   void train(const FluidDataSet<std::string, double, 1>& dataset, index k,
-             index maxIter)
+             index maxIter, unsigned initialize )
   {
     using namespace Eigen;
     using namespace _impl;
@@ -41,14 +50,14 @@ public:
     {
       mK = k;
       mDims = dataset.pointSize();
-      initMeans(dataPoints);
+      initMeans(dataPoints, initialize);
     }
 
     while (maxIter-- > 0)
     {
       mEmbedding = mMeans.matrix() * dataPointsT;
       auto assignments = assignClusters(mEmbedding);
-      if (!changed(assignments)) { break; }
+      if (mAssignments.rows() && !changed(assignments)) { break; }
       else
         mAssignments = assignments;
       updateEmbedding();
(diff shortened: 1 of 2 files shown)


tremblap:
SC and docs added:

https://github.com/flucoma/flucoma-sc/pull/178
https://github.com/flucoma/flucoma-docs/pull/205

weefuzzy:
You were underwhelmed with the improvement from Forgy initialisation on the 'rays' test dataset. Like random partition, just selecting k data points *can still be arbitrarily bad*, and how bad each of them is partly a function of the data distribution. 

Following images are plots of 10 runs of random partition, Forgy, kmeans++ and the mcmc approximation against the rays data both centred and un-centred. The coloured points in each plot are the estimated centroids, and the grey points are the normalised data. Hopefully this sheds some light on how these different schemes compare. 

Take home is: kmeans++ performs consistently very well, as expected. MCMC does almost as well (for an order of complexity less). Random partition is pretty bad, especially on the non-centred case. Forgy is pretty variable: it often does ok, but can do appallingly sometimes; in the centred case it's more prone to missing one of the clusters entirely, leaving much more work for the subsequent iteration to do. 

#### Centred

rays_centroids_centered

#### Not centred

rays_centroids_uncentered

tremblap:
this is an amazing explanation and demonstration. Thank you for taking the time.

tremblap:
Another question: I find `initialize` to be ambiguous.... I feel the verb implying a bolean in our interface. In effect the full sentence would be `initialisation method` so i think `initMethod` might be more instructive to the musician reader. it gets me to think a bit about what is under the hood. then I could write a one liner for each of the available methods (we keep) to give the pros and cons.

tremblap:
the branch is called keans-init - that might make the cross repo compile script bail.

tremblap:
ok , FluidManipulation doesn't build. My chain has been updated yesterday (bloody apple) and there are so many warnings that it fills all of my terminal buffer and I cannot find the error...

tremblap:
(building FluidManipulation from the tip of main is producing just the usual 3 warnings "unused parameter 'c'" and "implicit conversion changes signedness: 'index' (aka 'long') to 'size_type' (aka 'unsigned long')" 

and 2 notes: " note: in instantiation of function template specialization 'fluid::client::dataset::sortedDistances<const float, const double>'..."

tremblap:
ok it seems that tip of main hadn't been merged in. sorry for the noise

tremblap:
ok the new version with .initMethod_(0) which is supposed to be the same as before still gets much worse results in KMeans.

With the following code (ommiting line 37 and 43 with the current nightlies to compare) we get 25 to 30 empty clusters in the new version, by opposition to 3 to 6 in the old one...

```
~ds = FluidDataSet(s).read(FluidFilesPath("../Data/rays.json"));

(
~ds.dump({
	arg dict;
	defer{
		~fp = FluidPlotter(dict:dict,xmin:-1,ymin:-1).pointSizeScale_(3);
	};
});
)

(
~ls = FluidLabelSet(s);
~skm = FluidSKMeans(s, maxIter: 1000);
)

// re-inits, graphically, 5 per second, do 100 and count the empty clusters
(
Routine{
	var num = 0;
	100.do{|i|
		i.postln;
		~skm.clear;
		~skm.fitPredict(~ds,~ls,{
			~ls.dump({
				arg dict;
				~fp.categories_(dict);
			})
		});
		0.2.wait;
	};
	"% empty clusters on 100\n".postf(num);
}.play;
)

// rerun above with other initialisation
~skm.initMethod_(0)


/////// with kmeans, same jam

~km = FluidKMeans(s, maxIter: 1000);
~km.initMethod_(0)

~ds = FluidDataSet(s).read(FluidFilesPath("../Data/gaussian4.json"));
~ls = FluidLabelSet(s);

(
~ds.dump({
	arg dict;
	defer{
		~fp = FluidPlotter(dict:dict).pointSizeScale_(3);
	};
});
)

(
Routine{
	var num = 0;
	100.do{|i|
		i.postln;
		~km.clear;
		~km.fitPredict(~ds,~ls,{|x|
			if (x.includes(0)){num = num + 1};
			~ls.dump({
				arg dict;
				~fp.categories_(dict);
			})
		});
		0.2.wait;
	};
	"% empty clusters on 100\n".postf(num);
}.play;
)
```

PR #288, opened 2024-12-17
fix #270 - memory allocation for the autocorrelation in transient*

### LINKED ISSUE #270: `fluid.transientslice~`: audio thread crashes at higher `order`
### Please tell us what you were doing! You can include code and files by drag and dropping them into the text area.

report via a user: sporadic crashes with transient slice, stack trace indicated that these were in the detection function. 

Can reproduce with `@order 196` simply by turning audio on in the following patch. 

<pre><code>
441.3ocsT1rSDCBDG+b6SAgil5lB8y0S9dXLFjhJltPCP0UMtO6VFZ0UypV0
5EZloCy+4Gy.OEGguTuUXwnSPmghhdJNJBb4cDMZGg2v1xaYVHLr3wFFeGNI
7KU+FopU3f+QGc1wb7ajpquvH3tPtyKnqRSPzpZ+m7hoUz4ukHcuaJSoidkM
fl5Ku83Zr20yww9kjYVrJw8C6cpXchsP4f800NT5phChQ1miAcc8phDDoL2C
PQIvD8SwfL5M3x8PmHjFrUdsh0heca6wY0Bx4Q+39DsFZPTRncs1S6BCX4uA
PKW2I90zrFngjk95m4O2ksf8C9C7VwNDIM8GSRUXvKqxiP4+wfGcAA8p1dYy
JmgorRgxYak7AtOUaZDFDYc4Awm78Mxb.eR3Ujufe52veBBevy.x9mAP.3Vo
5iOPBEn2+6OXr5dCeRnP1RPuUgMBqSpXNoVsWLYgXN3Q+b0obF5PV.cplgN9
Zg7G0gNScdOOg1Eqq6NgwNFLHwvD6sZi2rNALkpfYAXZD2ImhGlavLyv7maX
3q2.kEdaYNNrU8v.rpWBiqwd3FjDtMnXaD1NVfC3RS7ywuvVYaKm
</code></pre>

Need to narrow down why. Possibilities include, in order of likelihood:  
* buffer overrun because we've messed up the allocation needed to account for blocksize vs order vs padding 
* screw up in the autocorrelation fn
* badness due to Eigen's solver deciding to allocate behind the scenes 

Not crashing

crashing

Mac

### Operating system version

13.5.2

### FluCoMa Version

1.0.6

tremblap:
allocates the maximum ARMmodel size so it fulfills the conditions of resizing (line 446)

### DIFF
--- a/include/algorithms/public/TransientExtraction.hpp
+++ b/include/algorithms/public/TransientExtraction.hpp
@@ -35,7 +35,7 @@ public:
 
   TransientExtraction(index maxOrder, index maxBlockSize, index maxPadSize,
       Allocator& alloc = FluidDefaultAllocator())
-      : mModel(maxOrder, maxBlockSize + (2 * maxPadSize), alloc),
+      : mModel(maxOrder, asUnsigned(maxBlockSize + (2 * maxPadSize) + maxOrder), alloc),
         mInput(asUnsigned(maxBlockSize + (2 * maxPadSize) + maxOrder), alloc),
         mDetect(asUnsigned(maxBlockSize), alloc),
         mForwardError(asUnsigned(maxBlockSize + maxOrder), alloc),


tremblap:
sadly this doesn't solve (as I hoped) #289 so there is now an issue there.

AlexHarker:
You will note that GitHub doesn't link issues mentioned in the title - it would be clearer to put things in the body text of the merge request. I would also advise against "this doesn't fix this issue" comments unless they undo an earlier comment on this merge request, as this just links out to an unrelated issue now.

> If the algo is right at line 446, then the requested size was wrong, so I corrected the latter.

I assume you mean in the file edited? That doesn't appear necessarily correct to me as a statement. See this line which sends only analysis size through to the model.

https://github.com/flucoma/flucoma-core/blob/8539bcd0d3186a2098be07f2b0638f5b851c6da6/include/algorithms/public/TransientExtraction.hpp#L156

If you are not verbose in your comments it is much harder to fulfil a review meaningfully. There's obviously an issue if an assert is firing, but at the moment I'm not sure where the issue is as there are at least 3 files involved, and if I understand correctly the assert might be in the FFT file.

tremblap:
I was hoping that this fix (which does fix #270 as long as we don't go up with any of the parameters after initialisation) would also fix #289 - it doesn't. Sorry if I was unclear. The current fix (#288) is valid for #270 but we need a way to cap the 3 interlocking values at runtime. I have not included that latter solution yet as I don't know if/how our parameter checks can do this.

AlexHarker:
Suggest to close this request as covered in #270 as not the correct fix for the issue.

weefuzzy:
Superseded by #290

PR #295, opened 2025-03-19
Introduces a basic test for MLP / SGD 

weefuzzy:
Adds a basic consistency test for MLP / SGD, ahead of doing some more substantive work to address some other issues. 

In course of doing this, found a simple bug (w/r/t Torch's calculations) that will probably have been slowing down convergence a bit (to do with the loss derivative that gets passed into backprop) 

Bug is a one-liner. Most of the code is the test and its gubbins

Not expecting the reviewers to have much to say about the code, but you can confirm (if you like) that your personal MLPs still converge etc.

### DIFF
--- a/include/algorithms/public/MLP.hpp
+++ b/include/algorithms/public/MLP.hpp
@@ -168,6 +168,7 @@ public:
     index    nRows = out.rows();
     ArrayXXd chain =
         ArrayXXd::Zero(nRows, mLayers[mLayers.size() - 1].inputSize());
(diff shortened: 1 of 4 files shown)


tremblap (approved):
It works both in my overfitting examples (a la Fiebrink) and in the autoencoder (8c) test, as well as the 2 'unit testers' so I'd merge now, and release to the community to see if Rod (and others :) ) find it converges faster (it felt like it here but it might just be my imagination)

PR #296, opened 2025-03-20
Ensures source and target IDs are paired up in SGD

### LINKED ISSUE #292: `MLP*`: Doesn't respect input IDs
The MLP client training code seems to assume that the input and output data / labels are identically ordered, and never checks that this is the case. Much of the time it is, but one can simply 'reorder' the entries of a Dataset – which should remain semantically equivalent because it's an unordered collection – and completely screw up one's training. 

There's no particularly nice solution to this: we just have to traverse the ID list of one of the data sets and rearrange (a copy of) the other to match. Current fix in progress introduces a new flag (`strict`) to enable this, because it's not cheap.

weefuzzy:
fixes #292 

Ok, ok, I know this *looks* like a massive PR, but it's not really 😄 

### The problem
`SGD::train` assumes that its source and target data are *aligned*, i.e. that each index represents the same ID. This not necessarily the case because we're using unordered containers by design. You can make training silently fail to do anything sensible by adding points to source and target in a different order, for instance. This is baaaaad.

### But why is this diff so large, you 'orrible man? 
The core of this PR is decoupling the batching and sampling of data from `SGD`, a bit like pytorch. I like that `SGD` doesn't want to know about `DataSet`, it just needed to go further. 

So we have 
* some new classes that do data sampling: a base class, an implementation that properly matches source and target IDs, and a simple implementation for when data really are aligned (this opens up some cool future extensions*). 
* changes to SGD to remove its batching code and use this instead
* supporting changes to `FluidDataSet`, `FluidTensorView` 
* some tests 
* updates to MLP clients to actually check that IDs match and use appropriate sampler 

User-side, nothing much should change besides a couple of error messages: so, reviewers: please test this proposition by training against your own stuff and making sure it still works as it did. 

\* Like, it would now be pretty easy to enable quicker n' dirtier MLP training (resp. the rest of the data objects) with CCE buffers alone for those cases where people don't want / care about IDs. I think this would have significant workflow and teaching upsides...

### DIFF
--- a/include/flucoma/algorithms/public/MLP.hpp
+++ b/include/flucoma/algorithms/public/MLP.hpp
@@ -30,8 +30,6 @@ class MLP
   using ArrayXXd = Eigen::ArrayXXd;
 
 public:
-  explicit MLP() = default;
-  ~MLP() = default;
 
   void init(index inputSize, index outputSize,
             FluidTensor<index, 1> hiddenSizes, index hiddenAct, index outputAct)
--- a/include/flucoma/algorithms/public/SGD.hpp
+++ b/include/flucoma/algorithms/public/SGD.hpp
@@ -15,10 +15,10 @@ under the European Union’s Horizon 2020 research and innovation programme
 #include "../../data/FluidDataSet.hpp"
 #include "../../data/FluidIndex.hpp"
 #include "../../data/FluidTensor.hpp"
+#include "../../data/SimpleDataSampler.hpp"
 #include "../../data/TensorTypes.hpp"
 #include <Eigen/Core>
 #include <limits>
-#include <random>
 
 namespace fluid {
 namespace algorithm {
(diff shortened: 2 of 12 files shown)


tremblap:
Maybe @lewardo wants to also try this in his spare time. That will have consequences for the WiP of DataSeries for which @weefuzzy has another idea IIRC but for now it would be good to also get him to read it. Also if @AlexHarker wants to try this bespoke objects for his piece from that branch that might be a good idea. I will run the test on Monday as I am in compo/impro/code residency for the next days...

tremblap:
it all works here. Is it possible that it doesn't bounce as much as before, or have I been lucky?

tremblap (approved):

tremblap:
I think I've been lucky. In all cases it works fine.

jamesb93:
Working for me!

PR #297, opened 2025-03-27
Update include layout

weefuzzy:
* Moves flucoma headers from `include` to `include/flucoma`
* Updates `#includes` accordingly for examples and test
* fixes a few incorrectly expressed includes in core headers
* updates cmake paths

🚨  To test this with CCEs you'll need the correspondingly named PR branches for each environment, otherwise it won't build


tremblap (approved):
it compiles fine

PR #299, opened 2025-03-31
Catch2: Migrate tests to v3

weefuzzy:
Overdue getting-with-the-times and adapting our tests to v3 of Catch2.


weefuzzy:
Test failure nothing to do with this (to do with CMake warning about JSON lib becoming an error in between runs!). Will fix that elsewhere

Issue #269, opened 2024-04-30
Bug: Sample Rate issue with `fluid.bufcompose~`


rconstanzo:
### Please tell us what you were doing! You can include code and files by drag and dropping them into the text area.

As outlined in [this thread on the discourse](https://discourse.flucoma.org/t/sample-rate-bug-in-fluid-bufcompose/2273), if the `buffer~` that `fluid.bufcompose~` is copying into has no size (as in, not declared, or after a `sizeinsamps 0`), when you copy something into it, the SR is set to the system SR, and not the `@source` SR or even the SR if the `buffer~` if you manually declare it.

This means that any automatic buffer management-related things with `fluid.bufcompose~` will end up getting the SRs changed for all the newly created buffers, which can be a huge problem on its own. And whenever using `sizeinsamps 0` to specifically empty/clear a `buffer~` for iterative/loop-based processes, the SRs also get changed.

As mentioned by @weefuzzy "we should probably check the behaviour for `bufnmf~` because that also has its own witchy buffer writing code"

![7ed66b122b67368758421f1a60a2825999de154d](https://github.com/flucoma/flucoma-max/assets/2747015/afa1f37f-6835-40c8-9228-5ab92bc252ed)
```
----------begin_max5_patcher----------
1666.3oc0ZszjiZCD9rmeEpnxosbbgdwibJImxgjKYOlJ0TXaYOrqMxEHlcl
r0t+1idAFx.iYLHSlspEvZDzp+Z0e8C3q2svaM+IVgG3m.+EXwhud2hE5gTC
rv96EdGSdZygjB8z71vOdjkI7VZ9aB1SB83Qq.oB.+QVdd5VVAP7.CHOKRyR
Do7LvG+ypa4PZFaCuLSeeH6fqKWu9.SMBr0HExGlZTe6n63YhcIaZMS0ZXC+
.O2nGvU9KA90Gj+D721YltUuV4q+zORIUqmrxioYGXhhlOSkbJR+GsbPpmS8
b4khpI2bQkkbTOYueic3QlHcSR0y+ThXyCoY6uOmsQXWhwnUzkf3X8ZEFGqN
EPUqT087s6tScX4HMIgq.ELQKq.eGXc4N4MbhWvdUCRW.cS3COivWDUgdQ9Z
XCFoOQBlXzKX8JvlGXa9rbqqB2XGOIdVgd6X4fj8IoYiB+PyH9E3GpfLJ1s.
[...]

PR #300, opened 2025-04-07
bufcompose now takes the source samplerate when resizing - fixes 269

tremblap:


tremblap:
@weefuzzy my PR in SC is about to be merged in (famous last words) so I think we should merge that in for #269 and @rconstanzo would be happy

weefuzzy (changes requested):

PR #301, opened 2025-04-07
Fix missing template args caught by new clang

weefuzzy:
@tremblap discovered that the most recent clang is newly upset by calls like `x.template f()`, i.e. where the template argument list is absent. We had two such, seemed like a resonable thing to complain about really.


tremblap:
it works perfectly thanks. I still get yellow warnings about "cmake --help-policy CMP0177" but hey it all rocks

tremblap (approved):

PR #302, opened 2025-04-11
Feature/add DataSetRead to FluidDataSet

tremblap:
for staying on server, mimicks the KDTree.kr behaviour. Only affect SuperCollider (but 2 bug fixes were undisciplinarily included, I expect the wrath of the gods)

### DIFF
--- a/include/flucoma/clients/nrt/DataSetClient.hpp
+++ b/include/flucoma/clients/nrt/DataSetClient.hpp
@@ -452,11 +452,12 @@ public:
 
       if (lookupDSpointer)
       {
+        auto lookupDS = lookupDSpointer->getDataSet();
+        auto inputDSids = inputDSpointer->getDataSet().getIds();
         for (index i = 0; i < mNumValidKs; i++)
         {
-          lookupDSpointer->getDataSet().get(
-              inputDSpointer->getDataSet().getIds()[indices[i]],
-              mRTBuffer(Slice(i * pointSize, pointSize)));
+          lookupDS.get(inputDSids[indices[i]],
+                       mRTBuffer(Slice(i * pointSize, pointSize)));
         }
       }
 
--- a/include/flucoma/clients/nrt/DataSetClient.hpp
+++ b/include/flucoma/clients/nrt/DataSetClient.hpp
@@ -374,7 +374,7 @@ public:
   static constexpr auto& getParameterDescriptors() { return DataSetReadParams; }
 
   DataSetRead(ParamSetViewType& p, FluidContext& c)
-      : mParams(p), mRTBuffer(c.allocator())
+      : mParams(p)
   {
     controlChannelsIn(1);
     controlChannelsOut({1, 1});
@@ -454,7 +454,6 @@ public:
   }
 
 private:
-  RealVector            mRTBuffer;
   index                 mNumValidKs = 0;
   InputDataSetClientRef mDataSetClient;
 };
--- a/include/flucoma/clients/nrt/DataSetClient.hpp
+++ b/include/flucoma/clients/nrt/DataSetClient.hpp
@@ -413,13 +413,12 @@ public:
                 // unavailable");
 
       auto outputDSpointer = get<kOutputDataSet>().get().lock();
-      if (!outputDSpointer)
-        return; // c.reportError("FluidDataSet RT Query could not obtain
-                // reference (output) FluidDataSet");  //TODO: we might as well
-                // consider passing the distances in this case (and in kdtree
-                // too!)
+      //      if (!outputDSpointer)
+      //        return; // c.reportError("FluidDataSet RT Query could not obtain
+      // reference (output) FluidDataSet");
 
-      index pointSize = outputDSpointer->dims();
+      index pointSize =
+          outputDSpointer ? (index) outputDSpointer->dims() : (index) 1;
 
       auto  outBuf = BufferAdaptor::Access(get<kOutputBuffer>().get());
       index maxK = outBuf.samps(0).size() / pointSize;
@@ -452,11 +451,22 @@ public:
 
       mNumValidKs = std::min(asSigned(indices.size()), maxK);
 
-      for (index i = 0; i < mNumValidKs; i++)
+      if (outputDSpointer)
       {
-        outputDSpointer->getDataSet().get(
-            inputDSpointer->getDataSet().getIds()[indices[i]],
-            mRTBuffer(Slice(i * pointSize, pointSize)));
+        for (index i = 0; i < mNumValidKs; i++)
+        {
+          outputDSpointer->getDataSet().get(
+              inputDSpointer->getDataSet().getIds()[indices[i]],
+              mRTBuffer(Slice(i * pointSize, pointSize)));
+        }
+      }
+
+      else
+      {
+        for (index i = 0; i < mNumValidKs; i++)
+        {
+          mRTBuffer[i] = distances[indices[i]];
+        }
       }
 
       outBuf.samps(0, outputSize, 0) <<= mRTBuffer;
--- a/include/flucoma/clients/nrt/DataSetClient.hpp
+++ b/include/flucoma/clients/nrt/DataSetClient.hpp
@@ -406,42 +406,30 @@ public:
                 // unavailable");
 
       auto lookupDSpointer = get<kLookupDataSet>().get().lock();
-      //      if (!lookupDSpointer)
-      //        return; // c.reportError("FluidDataSet RT Query could not obtain
-      // reference (output) FluidDataSet");
 
       index pointSize = lookupDSpointer ? lookupDSpointer->dims().value() : 1;
 
-      auto  outBuf = BufferAdaptor::Access(get<kOutputBuffer>().get());
-      index maxK = outBuf.samps(0).size() / pointSize;
-      if (maxK <= 0) return;
-      index outputSize = maxK * pointSize;
+      auto outBuf = BufferAdaptor::Access(get<kOutputBuffer>().get());
+      mNumValidKs = outBuf.samps(0).size() / pointSize;
+      if (mNumValidKs <= 0)
+      {
+        output[0](0) = 0;
+        return;
+      }
+      index outputSize = mNumValidKs * pointSize;
 
       RealVector point(dims, c.allocator());
       point <<= BufferAdaptor::ReadAccess(get<kInputBuffer>().get())
                     .samps(0, dims, 0);
+
       if (mRTBuffer.size() != outputSize)
       {
         mRTBuffer = RealVector(outputSize, c.allocator());
         mRTBuffer.fill(0);
       }
 
-      std::vector<index> indices(asUnsigned(inputDSpointer->size()));
-      std::iota(indices.begin(), indices.end(), 0);
-      std::vector<double> distances(asUnsigned(inputDSpointer->size()));
-
       auto inputdata = inputDSpointer->getDataSet().getData();
-
-      std::transform(indices.begin(), indices.end(), distances.begin(),
-                     [&point, &inputdata, this](index i) {
-                       return distance(point, inputdata.row(i));
-                     });
-
-      std::sort(indices.begin(), indices.end(), [&distances](index a, index b) {
-        return distances[asUnsigned(a)] < distances[asUnsigned(b)];
-      });
-
-      mNumValidKs = std::min(asSigned(indices.size()), maxK);
+      auto distances = sortedDistances(point, inputdata, c.allocator());
 
       if (lookupDSpointer)
       {
@@ -449,7 +437,7 @@ public:
         auto inputDSids = inputDSpointer->getDataSet().getIds();
         for (index i = 0; i < mNumValidKs; i++)
         {
-          lookupDS.get(inputDSids[indices[i]],
+          lookupDS.get(inputDSids[distances[i].first],
                        mRTBuffer(Slice(i * pointSize, pointSize)));
         }
       }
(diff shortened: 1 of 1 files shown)


tremblap:
hmmm. before I do all this, I will push now the new feature (which should have been in KDTree too, which I will also implement there after).

the new feature is simple: for this, if we don't pass a reference to a lookupDataSet, nothing comes out. In KDTree, it defaults to the underlying DS. In both cases, it is a lost opportunity to pass on distances. In a few cases, I'd like to know the distance to the K nearest on the server and this next item does that elegantly.

include/flucoma/clients/nrt/DataSetClient.hpp
```diff
@@ -351,8 +351,140 @@ class DataSetClient : public FluidBaseClient,
...
+        return; // c.reportError("FluidDataSet RT Query i/o buffers are
+                // unavailable");
+
+      auto outputDSpointer = get<kOutputDataSet>().get().lock();
+      if (!outputDSpointer)
+        return; // c.reportError("FluidDataSet RT Query could not obtain
+                // reference (output) FluidDataSet");  //TODO: we might as well
+                // consider passing the distances in this case (and in kdtree
+                // too!)
+
```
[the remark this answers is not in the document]
tremblap: see the new push with the added feature

tremblap:
btw, I reckon my next homework will be to do the same to KDTree... since most of the code is taken from there it should be faster while I have my head in all of it?

tedmoore:
Getting the distances is great! 

What's an example of a use case where I would want to do a KNN search on a FluidDataSet? Why not just use the KDTree? When/what are the chances that a brute force will be more efficient/faster than a KDTree?

tremblap:
I'm happy distance is great to you too :)

I use brute force when I do search on circular buffers that I write at the same time. THis is sexy as hell, easy to do in Max but hard to do in SC on the server - KDTree build is slow and needs to do a sort at least so building every query makes no sense for a small dataset. So that this the use case - I use it now in the language but that is back/forth/back/forth for nothing so I am trying to do it server side. It is also a way for me to get less worse at c++ and to continue to understand the wrappers (as I am a bit stalled with the other bits I'm working on). Moreover, @weefuzzy is harsh on me for good reasons, and I learn A LOT!

tedmoore:
> I use brute force when I do search on circular buffers that I write at the same time.

Nice. Yes, this is a great use case!

include/flucoma/clients/nrt/DataSetClient.hpp
```diff
@@ -413,13 +413,12 @@ class DataSetRead : public FluidBaseClient, ControlIn, ControlOut
...
-                // reference (output) FluidDataSet");  //TODO: we might as well
-                // consider passing the distances in this case (and in kdtree
-                // too!)
+      //      if (!outputDSpointer)
+      //        return; // c.reportError("FluidDataSet RT Query could not obtain
+      // reference (output) FluidDataSet");
 
-      index pointSize = outputDSpointer->dims();
+      index pointSize =
+          outputDSpointer ? (index) outputDSpointer->dims() : (index) 1;
```
weefuzzy, line 421: No need for casts here. `dims()` is actually returning a `MessageResult<index>` (for consumption by the host), just call `value()` on that:
```suggestion
         outputDSpointer ? outputDSpointer->dims().value() : 1;
```
(And when we do cast in C++, bear in mind that C-style casts are rarely the thing you want: here `static_cast<T>` would have been the thing)

include/flucoma/clients/nrt/DataSetClient.hpp
```diff
@@ -351,8 +351,150 @@ class DataSetClient : public FluidBaseClient,
...
+constexpr auto DataSetReadParams = defineParameters(
+    InputDataSetClientRef::makeParam("dataSet", "DataSet Name"),
+    LongParam("numNeighbours", "Number of Nearest Neighbours", 1),
+    InputDataSetClientRef::makeParam("outputDataSet", "Output DataSet Name"),
+    InputBufferParam("inputPointBuffer", "Input Point Buffer"),
+    BufferParam("predictionBuffer", "Prediction Buffer"));
+
+class DataSetRead : public FluidBaseClient, ControlIn, ControlOut
+{
+  enum { kDataSet, kNumNeighbors, kOutputDataSet, kInputBuffer, kOutputBuffer };
```
weefuzzy, line 365:
```suggestion
enum { kDataSet, kNumNeighbors, kLookupDataSet, kInputBuffer, kOutputBuffer };
```

include/flucoma/clients/nrt/DataSetClient.hpp
```diff
@@ -351,8 +351,150 @@ class DataSetClient : public FluidBaseClient,
...
+
+      index dims = inputDSpointer->dims();
+
+      InOutBuffersCheck bufCheck(dims);
+      if (!bufCheck.checkInputs(get<kInputBuffer>().get(),
+                                get<kOutputBuffer>().get()))
+        return; // c.reportError("FluidDataSet RT Query i/o buffers are
+                // unavailable");
+
+      auto outputDSpointer = get<kOutputDataSet>().get().lock();
```
weefuzzy, line 415:
```suggestion
      auto lookupDSpointer = get<kLookupDataSet>().get().lock();
```

include/flucoma/clients/nrt/DataSetClient.hpp
```diff
@@ -27,6 +27,28 @@ enum { kName };
 constexpr auto DataSetParams =
     defineParameters(StringParam<Fixed<true>>("name", "Name of the DataSet"));
 
+double distance(FluidTensorView<const double, 1> point1,
+                FluidTensorView<const double, 1> point2)
```
weefuzzy, line 31: This will let you get rid of the temporary buffer for the input point

```suggestion
template<typename T, typename U> 
double distance(FluidTensorView<T, 1> point1,
                FluidTensorView<U, 1> point2)
```

include/flucoma/clients/nrt/DataSetClient.hpp
```diff
@@ -27,6 +27,28 @@ enum { kName };
...
+{
+  return std::transform_reduce(
+      point1.begin(), point1.end(), point2.begin(), 0.0, std::plus{},
+      [](double v1, double v2) { return (v1 - v2) * (v1 - v2); });
+};
+
+auto sortedDistances(FluidTensorView<const double, 1> x,
+                     FluidTensorView<const double, 2> Y, Allocator& alloc)
```
weefuzzy, line 39:
```suggestion
template<typename T, typename U> 
auto sortedDistances(FluidTensorView<T, 1> x,
                     FluidTensorView<U, 2> Y, Allocator& alloc)
```

include/flucoma/clients/nrt/DataSetClient.hpp
```diff
@@ -223,33 +244,23 @@ class DataSetClient : public FluidBaseClient,
     FluidTensor<const double, 1> point(
         BufferAdaptor::ReadAccess(data.get()).samps(0, mAlgorithm.dims(), 0));
```
weefuzzy, line 245:
```suggestion
 auto point =  
          BufferAdaptor::ReadAccess(data.get()).samps(0, mAlgorithm.dims(), 0);  
```

include/flucoma/clients/nrt/DataSetClient.hpp
```diff
@@ -266,42 +277,31 @@ class DataSetClient : public FluidBaseClient,
...
```
weefuzzy, line 278:
```suggestion
auto point =
        BufferAdaptor::ReadAccess(data.get()).samps(0, mAlgorithm.dims(), 0);
```

include/flucoma/clients/nrt/DataSetClient.hpp
```diff
@@ -337,22 +337,132 @@ class DataSetClient : public FluidBaseClient,
...
+      if (mNumValidKs <= 0)
+      {
+        output[0](0) = 0;
+        return;
+      }
+      index outputSize = mNumValidKs * pointSize;
+
+      RealVector point(dims, c.allocator());
+      point <<= BufferAdaptor::ReadAccess(get<kInputBuffer>().get())
+                    .samps(0, dims, 0);
```
weefuzzy, line 423:
```suggestion
      auto outSamps = outBuf.samps(0);
    
      index numPoints = outSamps.size() / pointSize;
      if (numPoints <= 0)
        return; //log "Output buffer 0-sized"
      
      auto point = BufferAdaptor::ReadAccess(get<kInputBuffer>().get()).samps(0, dims, 0);
```

include/flucoma/clients/nrt/DataSetClient.hpp
```diff
@@ -337,22 +337,132 @@ class DataSetClient : public FluidBaseClient,
...
+
+      auto inputdata = inputDSpointer->getDataSet().getData();
+      auto distances = sortedDistances(point, inputdata, c.allocator());
+      auto outSamps = outBuf.samps(0);
```
weefuzzy, line 427:
```suggestion
```

include/flucoma/clients/nrt/DataSetClient.hpp
```diff
@@ -337,22 +337,132 @@ class DataSetClient : public FluidBaseClient,
...
+        std::for_each_n(distances.begin(), mNumValidKs,
+                        [pointSize, &outSamps, &lookupDS, &inputDSids,
+                         n = 0](auto& p) mutable {
+                          if (auto point = lookupDS.get(inputDSids[p.first]);
+                              point.data() != nullptr)
+                          {
+                            outSamps(Slice(n, pointSize)) <<= point;
+                          }
+                          n += pointSize;
+                        });
```
weefuzzy, line 442: I poked around for a bit to try and make this nicer. I think the best we can do for the moment is separate the lambda out from the `for_each` call to help readability. 

```suggestion 
        auto lookupDS = lookupDSpointer->getDataSet();
        auto ids = inputDSpointer->getDataSet().getIds();
        
        auto lookupFn = [&lookupDS, ids, outSamps, pointSize,
                         n = 0](auto p) mutable {
          auto id = ids[p.first];
          if (auto point = lookupDS.get(id); point.data() != nullptr)
            outSamps(Slice(n, pointSize)) <<= point;
          n += pointSize;
        };

        std::for_each_n(distances.begin(), numPoints, lookupFn);
```

include/flucoma/clients/nrt/DataSetClient.hpp
```diff
@@ -337,22 +337,132 @@ class DataSetClient : public FluidBaseClient,
...
+      }
+      else
+      {
+        std::for_each_n(distances.begin(), mNumValidKs,
+                        [&outSamps, n = 0](auto& p) mutable {
+                          outSamps[n++] = std::sqrt(p.second);
+                        });
```
weefuzzy, line 449: This can be a transform

```suggestion 
        std::transform(distances.begin(), distances.begin() + numPoints,
                       outSamps.begin(),
                       [](auto p) { return std::sqrt(p.second); });
```

include/flucoma/clients/nrt/DataSetClient.hpp
```diff
@@ -337,22 +337,132 @@ class DataSetClient : public FluidBaseClient,
...
+      }
+    }
+
+    output[0](0) = mNumValidKs;
```
weefuzzy, line 453:
```suggestion
      mLastNumPoints = numPoints;       
    }
    output[0](0) = mLastNumPoints; //re-report last output in absence of trigger
```

include/flucoma/clients/nrt/DataSetClient.hpp
```diff
@@ -337,22 +337,132 @@ class DataSetClient : public FluidBaseClient,
...
+  }
+
+private:
+  index                 mNumValidKs = 0;
```
weefuzzy, line 457:
```suggestion
  index mLastNumPoints{0}; 
```

include/flucoma/clients/nrt/DataSetClient.hpp
tremblap: is this really a double that is yield from distance, or we do auto like below?

tremblap: also, I sort of get the syntax of template definition first but a little verbose helper would help me understand the consequences to line 244 - I like the elegance but I want to understand the grammar and logic, please

weefuzzy: Templating on the value types of the `FluidTensorView`s means that we're no longer need to construct temporary buffers of `FluidTensor<double,N>` just to do conversion from `float` – that can happen in `transform_reduce` instead. So on line 244 (and its counterparts) instead of making a new `FluidTensor` we can just take the `FluidTensorView<float,1>` returned by `samples()` and use that directly.

tremblap: ok I've implemented it all and pushed. but I still have a question about the templating. SHould we not force the templates to have the same typename T for both inputs to the function? or we just go super permissing like this and expect the first compile pass to check that T and U can actually sum to the difference of 2 vectors and generate double out of that?

weefuzzy: The point is that we need to be able to handle them being different (specifically `float` in one, `double` in other). We could add some (super-ugly before C++20) restrictions to insist that both `T` and `U` are at least numeric, or even floating point, which would be worth it if this function were made to be called more widely. 

For our actual use cases, there's no problem: `transform_reduce` will be using `double` internally (inferred from the type of the initialization value we pass), and if the compiler can't work out how to do `x - y` for each of the points, it will moan.

PR #303, opened 2025-04-16
Update kdtree kr with dist (and refactor)

tremblap:
Using the latest refactor in DS.kr, enacted the following changes:

1. made the variable names more explicit (mLastNumPoints and kLookupDataSet)
2. removed the temp output vector and its allocator
3. replaced loops with sdt::iterators
4. added the distance output when no lookupDataSet is provided

still 2 questions pending:
1. line 230: I wasn't able to remove the 'point' allocation, as the first argument of kdtree.algo.knearest is not templatable for me
2. line 243: I needed to dereference id - i don't know why, as I though that id would be an iteration of ids, each being a pointer to a string in the vector of strings pointers, like in DataSet.kr

 any pointer (pun intended) welcome

### DIFF
--- a/include/flucoma/clients/nrt/KDTreeClient.hpp
+++ b/include/flucoma/clients/nrt/KDTreeClient.hpp
@@ -78,22 +78,12 @@ public:
   MessageResult<StringVector> kNearest(InputBufferPtr  data,
                                        Optional<index> nNeighbours) const
   {
-    // we can deprecate ancillary parameters in favour of optional args by
-    // falling back to using parameters when arg not present
     index k = nNeighbours ? nNeighbours.value() : get<kNumNeighbors>();
-    // alternatively we could just be hardcore and ignore parameters and have
-    // message handlers fallback to a default when arg missing (which would be
-    // eventual behaviour, I guess) index k =  nNeighbours.value_or(1);
-    if (k > mAlgorithm.size()) return Error<StringVector>(SmallDataSet);
-    // if (k <= 0 && get<kRadius>() <= 0) return Error<StringVector>(SmallK);
-    if (!mAlgorithm.initialized()) return Error<StringVector>(NoDataFitted);
-    InBufferCheck bufCheck(mAlgorithm.dims());
-    if (!bufCheck.checkInputs(data.get()))
-      return Error<StringVector>(bufCheck.error());
-    RealVector point(mAlgorithm.dims());
-    point <<=
-        BufferAdaptor::ReadAccess(data.get()).samps(0, mAlgorithm.dims(), 0);
-    auto [dists, ids] = mAlgorithm.kNearest(point, k, get<kRadius>());
+
+    auto reply = computeKnearest(data, k);
+    auto dists = reply.value().first;
+    auto ids = reply.value().second;
+
     StringVector result(asSigned(ids.size()));
     std::transform(ids.cbegin(), ids.cend(), result.begin(),
                    [](const std::string* x) {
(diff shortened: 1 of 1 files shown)


tremblap:
the 2nd commit also corrects the error for non-triggered kr output post negative triggered output

weefuzzy:
> * line 230: I wasn't able to remove the 'point' allocation, as the first argument of kdtree.algo.knearest is not templatable for me

It's not calling into already generic code, so there would be some extra work in updating other functions too. Given the amount of other apparatus querying `KDTree` creates, the pay off from avoiding this one temporary pales a bit.

> * line 243: I needed to dereference id - i don't know why, as I though that id would be an iteration of ids, each being a pointer to a string in the vector of strings pointers, like in DataSet.kr

There's no particular reason to suppose it should work like the query for `DataSet`. `kNearest` here returns a `KNNResult` which is a `pair<vector<double>, vector<const string*>>`. So you need to deference the `id` when you call `get` because it's a `string*` and you need a `string&` there.

include/flucoma/clients/nrt/KDTreeClient.hpp
```diff
@@ -188,74 +194,74 @@ class KDTreeQuery : public FluidBaseClient, ControlIn, ControlOut
...
+      else
       {
-        dataset.get(*ids[asUnsigned(i)],
-                    mRTBuffer(Slice(i * pointSize, pointSize)));
+        std::transform(dists.begin(), dists.begin() + numPoints,
+                       outSamps.begin(), [](auto p) { return std::sqrt(p); });
       }
-      outBuf.samps(0, outputSize, 0) <<= mRTBuffer;
+
+      mLastNumPoints = ids.size();
```
weefuzzy, line 256: To be consistent with the old code, shouldn't this be `min(ids.size(), numPoints)`?

tremblap: indeed, sorted

tremblap:
thanks for this, pushed the change. Now, I have avoided the TODO at line 108, mostly because I didn't know how to overload a single function with a fork in the end... or maybe I have a private (computeKnearest that is called with a bool argument on output one or the other?)

tremblap:
I have 2 c++ potential improvments to be checked by @weefuzzy here.
1. line 250-251 - no need to square there, so I reckon I could just make a slice but I couldn't find the right syntax
2. line 84-85 seem to be the only way I managed to compile - otherwise I got the ugly error of <Cannot decompose class type 'MessageResult<std::pair<std::vector<double,...'

happy to learn again - or to merge as is because life is too short

include/flucoma/clients/nrt/KDTreeClient.hpp
```diff
@@ -105,19 +97,9 @@ class KDTreeClient : public FluidBaseClient,
...
-    if (!mAlgorithm.initialized()) return Error<RealVector>(NoDataFitted);
-    InBufferCheck bufCheck(mAlgorithm.dims());
-    if (!bufCheck.checkInputs(data.get()))
-      return Error<RealVector>(bufCheck.error());
-    RealVector point(mAlgorithm.dims());
-    point <<=
-        BufferAdaptor::ReadAccess(data.get()).samps(0, mAlgorithm.dims(), 0);
-    auto [dist, ids] = mAlgorithm.kNearest(point, k, get<kRadius>());
-    return {dist};
+    auto  reply = computeKnearest(data, k);
```
weefuzzy, line 101: You need to check that the result is `ok` here as well

weefuzzy (changes requested):

include/flucoma/clients/nrt/KDTreeClient.hpp
tremblap: oupsy, sorry, added now running the tests

PR #304, opened 2025-04-16
correct the initialisation post-trigger of the mLastNumPoints

tremblap:
we should make sure that post-trigger, the past is forgotten.


tremblap:
as per my pre-report:  inside the successful if line 391, we should set it to 0 there too. it means 'i had a query' and if it bails for any reason the user need to know so it should output 0 then, and all subsequent kr. in the current code, it sends 0 on the trigger process cycle, then all subsequent kr will send the last number of successful Ks which is wrong for me. so we should change the line to output[0](0) = mLastNumPoints = 0;

PR #305, opened 2025-05-23
dataset:tobuffer: adds a flag to flip the ID-label 

### LINKED ISSUE #276: Feature Request: tobuffer should save IDs to samples, not the other way around
### Is your feature request related to a problem? Please describe the problem.

When using `tobuffer` on a `fluid.dataset~`, you specify a `fluid.labelset~` that will contain the dataset ID (as the label) for each sample in the buffer. Then you can *only* query with a buffer sample index to get the corresponding dataset ID. I almost never need it this way, but I **always** need the opposite: to query with a dataset ID and get the corresponding buffer sample index. It is possible to work around this via dataset/labelset dumping, but that gets *very* prohibitive with large datasets.

### Describe the solution you'd like to see.

If `tobuffer` would create a labelset where the dataset ID becomes the labelset identifier and the buffer sample index the label, it would be much easier down the line to query which dataset ID got mapped to which index (without excessive loops).

### Describe alternatives you've considered

An alternative could be that `fluid.labelset~` gets an `indexof` method, which gets the first index (identifier) for a given label. But I don't think this would be as efficient as the above solution.

Another option (if this is not that expensive) to export 2 labelsets where in one we have `<buffer-index> : <dataset identifier>` (like now) and another where we have `<dataset identifier> : <buffer-index>` (which would be new). This option could mean that the new version wouldn't break backwards compatibility so much, since the additional labelset would be just the next element in the list of what the `dataset~` reports after `tobuffer`.

### Additional context

Normally, other fluid objects (such as `kdtree~`) will operate on the dataset, and likely give you back dataset IDs. Example:
`fluid.jit.plotter`: to efficiently create a mesh from the 2D dataset, I go `refer <datasetname>`--> `tobuffer` --> to matrix with `jit.buffer~`. Luckily this is super fast even with millions of points because I get to avoid loops. 

But if I am highlighting the dataset elements closest to the mouse pointer using a `kdtree~`, now I get dataset IDs which I need to map to buffer indices (which is the same as matrix indices) to know which points to "highlight". And for this it is unavoidable to dump the samples-to-ids labelset *at least once*, which can cause seconds of hanging with large datasets. It also adds the burden of now having to update two books with the same data relationship.

tremblap:
fixes #276 

Adds a flag after the labelset name when`tobuffer`-ing - enable searching for an ID and retrieve the channel or frame (depending on the transposition)

I know I need to update docs and all help, but before, this is a proof of concept being tested. @weefuzzy do you agree with its implementation?

### DIFF
--- a/include/flucoma/clients/nrt/DataSetClient.hpp
+++ b/include/flucoma/clients/nrt/DataSetClient.hpp
@@ -203,7 +203,7 @@ public:
   }
 
   MessageResult<void> toBuffer(BufferPtr data, bool transpose,
-                               LabelSetClientRef labels)
+                               LabelSetClientRef labels, bool flip)
   {
     if (!data) return Error(NoBuffer);
     BufferAdaptor::Access buf(data.get());
@@ -217,7 +217,7 @@ public:
                   : FluidTensorView<const double, 2>(mAlgorithm.getData())
                         .transpose();
     auto labelsPtr = labels.get().lock();
-    if (labelsPtr) labelsPtr->setLabelSet(getIdsLabelSet());
+    if (labelsPtr) labelsPtr->setLabelSet(getIdsLabelSet(flip));
     return OK();
   }
 
@@ -225,7 +225,7 @@ public:
   {
     auto destPtr = dest.get().lock();
     if (!destPtr) return Error(NoDataSet);
-    destPtr->setLabelSet(getIdsLabelSet());
+    destPtr->setLabelSet(getIdsLabelSet(false));
     return OK();
   }
 
@@ -329,13 +329,19 @@ public:
   }
 
 private:
-  LabelSet getIdsLabelSet()
-  {
-    algorithm::DataSetIdSequence seq("", 0, 0);
-    FluidTensor<string, 1>       newIds(mAlgorithm.size());
-    FluidTensor<string, 2>       labels(mAlgorithm.size(), 1);
-    labels.col(0) <<= mAlgorithm.getIds();
-    seq.generate(newIds);
+  LabelSet getIdsLabelSet(bool flip)
+    {
+      algorithm::DataSetIdSequence seq("", 0, 0);
+      FluidTensor<string, 1>       newIds(mAlgorithm.size());
+      FluidTensor<string, 2>       labels(mAlgorithm.size(), 1);
+      
+      if (flip) {
+          seq.generate(labels.col(0));
+          newIds <<= mAlgorithm.getIds();
+      } else {
+              labels.col(0) <<= mAlgorithm.getIds();
+              seq.generate(newIds);
+      }
     return LabelSet(newIds, labels);
   };
 };


tremblap:
@balintlaczko enthusiasm and use case for max is compelling here:
https://github.com/flucoma/flucoma-core/issues/276#issuecomment-2905047142

weefuzzy:
Hmm. I think I need convincing that adding a flag argument to `tobuffer` – which introduces some mystery, and tethers this ability to only this context – is actually better than just adding a new message to `LabelSet` to do the flipping ex post – whih introduces some little extra work in the CCE but has much less bounded scope for use.

tremblap:
thanks! the idea of a general swapping in labelset makes probably more sense as it is more versatile. @balintlaczko would that be a huge irritant?

balintlaczko:
No, not at all, it actually makes sense to have a general way of flipping it in the labelset! Although, I still see the logic in having the flag in `tobuffer` as well, and I think @weefuzzy's logic could as well apply to the existing flag: why do it in the dataset method, where you could transpose the buffer instead (which is the more general use case) in post? It is up to you two, but I think keeping the flag would smoothen the usage experience (like the built-in buffer transposition already smooths it). And there is an existing logic in the interface with the samps/chans flag (I assume you don't intend to remove that one). I understand the argument about tethering but I don't see the mystery.

Nevertheless, having the generic flip built-in to the labelset does of course make a lot of sense, and (a bit off-topic, but) I also found myself making a `fluid.buftranspose` abstraction a while ago, and I agree, the generic tool over time will cover more ground.

I think the best answer is to have both, I do see the logic and value of having in the dataset `tobuffer` method, but I can very well live with the labelset flip too! Your choice. :)

tremblap:
before I get coding a PoC what would be a good method name for that? swapIdsAndLabels?

balintlaczko:
Could it be an attribute/flag, something like `@column` and then the options are `ids` or `labels`?

weefuzzy:
> before I get coding a PoC what would be a good method name for that? swapIdsAndLabels?

`flip` or `swap`? 

I think an argument against an attribute (presumably on `LabelSet` is that changing this property means re-indexing the object, which might be a surprising thing for an attribute change to do.

balintlaczko:
True, and I like `flip`!

tremblap:
except if @weefuzzy has a clever way to iterate through all items and reject/overwrite doubles?

tremblap:
I had an epiphany: using some stl magic to find if the vector of labels is made of unique elements. It seems the wisdom is to stl::sort it then check for std::ajacent_find - although since I want to exit at the first find, an actual for-loop is probably more efficient... I'll implement that to try and be prepared to be chastised.

tremblap:
ok I almost got it work - I couldn't get the relational operators == to work without doing explicit referencing to strings... and it seems I'm comparing pointers now... I'm sure @weefuzzy will be appalled. I also struggle a lot to pass from FluidTensors to the world of vectors, so maybe I need a cheat sheet somehow...

TODO: also, `swapIdsAndLabels` is ugly as hell, especially in max. see the attached patch - to test and to see...

tremblap:
<pre><code>
1777.3ocwZ0sjhpCD9ZmmhTV0d2LVjHHvb244XqsrBRTyN.ghDbF2s1yy9I+
ABpfnfm8Bwh7a+0cmuNoC+9kYyiXeQ3yAuC9NX1re+xrY5hTELy99r4o3u1j
f45lMOi7IK5mye0Tkf7kPWr.DAhpJMqLklkPD5d.OUHqTTUJxVJMV2c4P9FD
U0eS6DGyIFIadDNa27Ws+C9gsY4Xwl8zrcqKHaDlVtzYgyq.HzS82Rj5IR9D
7CUO9yKund753P5FVZDMi.3f2RvQjjqBZzfAsSmfVA3dAKzQCSjGT8W3pQg1
c7OoxY.MN33MuVbKvoDAoXMICGknwjyCiTjAovfPscMrx5V2GZltkvG.3oDN
GuibgcNuPMnCWa.uTarpaaau1UjyBuWAqPtOOu3sIkz3EZGXNQ7ufXBWLtku
tivQFZ.r+RsU1IXTHtKCJ+SbNMliyhMvtaLOXC7xGz.6noo77LrUnvQgWIgT
J4jqZMdSXrO.zLvdRAAHX.NQ92dBXKqHEeOFamqPa4NumUqRnHMmtFTBWoci
k9zSpCbLcy8X77tBFfcZ8TCNkkgKNp8ca9qW6puFwVf6ocpQMoo33Cj30xIW
1k0XgnfFUJLwemUC+YyIoQj3FZ9YySH6vaN1vD0KCayJSw440UqqUq+mHqf.
DWllCRFG0QP2FB4neS0tmluv0rLxDLbpoKyAERdCV5aRNiMEzbA8.omv+vgR
f.ePFDTfIZumItngHokqltSjhycs1RSHGHEbo2cKuIoWRihm0nKJEzOY5AJ3
05hnYlhVUWTA4.sp+d0khKjhtPJ2kEZfM+qUUTGpggESJxJo0wt0lJqHoMJY
R2XdNdioyJaWU0Mr+glvFZUh6JCqSn+IkgznuKgs4iyVTwxIYzr7BBWRehEV
Yut5XxVbYhX8VVlfS+kVBfpfwWo9sVI7pUpvfV7+mBJNoF.6JnwrLkPzxRnJ
tZ5jTJVCbSvnaQFN+Jc1PszQkbIHK4Q3BkgxRIfppTvXIsqBdh7YqvVcNMK6
Lsnfk2ckEzc66ouQLYko8M15Z3qKyL0tV5SHVqnQa2NbRhcka6g+KbFUFriH
nFS.xotRCq4d4xYVRRK7Zp4vUpIV5iug7IMVrWOQMcFr7BFmn40V4X5N01LZ
Ul.ui2tDt3nQo2nnxH6Z30BRZdhDEsafb0AkK36YexsMrxQqoB3zo6ZtltIM
Xqx6iN7hS70r7qw+0MG3E7fMGpqe7uZ++N2Tk4XA0TgMiBbJt2rpnASr53WI
KxD6AvN0InN0InNzIN8pTZEVr6cVaNOr9OW++BJE4xBbphQYxzKgSkZIP+bU
3eAsRAYGHiQ4GAXgbelI.yyHMeIHgUFC3rsBvNYjoDBnX+Qw9T5FPt5L5GA6
JXrCDUyEpp2KoH2OcJ3foRAa1Klq6yYwnFgchZm6lBpeNnaBYjid4UH5DKzR
mFgs0aowdTsl73lzWDS9pwlflD8iABO.GsSGJHz7AQ1XnYlRMvos9jPy5Jfl
VfT0ecUCmUVroxXZoQ.skMU9.nY0aD76mngAC0y8dEhfAJDgOQYP43OLMwyT
HbFpPfdlRw.EhfNkAagUmhYnm3+t19n7rLQ3jy1y201c4iebeUBPVbfR9bjo
IxInm7D4tp4oWsIOwyYhSTTgTDI.ShDFWl5GQFNQdnE9dRzgbt1dNlpjwPAz
QdGLNcmPBc5vM+0GTC7To8xlpdT3obfMc.kqyM+V.M9suQeG7M93xhKzA8XY
gw.UTf1MNz6YXSK+EEb5NhF.9VdI97cG1EqM.iq890L4Xw8ojpMM0SNdyGfX
r.+NXCKg+N7pZfkC0B6OnT81aNF80qeWZytD54Ac4d6KyGWVE8clJ3ZoqFGZ
kQ7jGuYbPxa0CcyhC7BiQlabxs54jl53BRNIKVdnNb73zAtOXxhsFS60RD.+
+39D2SRxqecgxqSgephkYTwmd7PvUZg.iEG5dU0fdDZeLCiV47sRZ0M8ui99
2HssMoz3blj30Ngq70Wdlu8JQWfpdqCy08HXvwHXggOGASEd5lRluSehl8p1
LWbrUzzuMEhF5VJM0dl.idpfCvwwcYqFwJhM2vC74O0va4zpz4KC0FBnu1UQ
+1kBqy3DVmAHr9vIvhrZHFjIXd7FxD07XuO9LMDJfkSzDcSD4LASzYKHt4o0
e3IZ4vo0G07.GzDoNvRCK4YKFQglyk3GXNWsIfu+3oCqbB6es2TrlPeX6AnG
B5iTBsBoYk77zeWQtd1u5lluAc8LWMhql3x0a7JIc3fasLytTbBlog4rLdTM
HaeuQoMmPdIrQJPN6kUMheqeazR88rZZbS0P3VctbsgYCmm84Wnljy9rKN6S
t3xO2ht+TKN+yrPmbxt97Jd4Ou7e.hit3EB
</code></pre>

tremblap:
one more thought: once I have sorted the column of the indices, can I trust it will still make sense to just copy the column of id? if not, then I need to find a way to sort the calling labelset by label, check each label is different to the next one, then do the copies in the destination...

tremblap:
ok here is the fat binary and the 'test code' - @balintlaczko what kind of demo code you felt like writing? Keep it super simple, as I have to code it in Pd and SC ;)

[fluid.libmanipulation.mxo.zip](https://github.com/user-attachments/files/20555087/fluid.libmanipulation.mxo.zip)

<pre><code>
1086.3oc0YksaaiCE8Y6uBBgoOMpBbQq4o9eLXPAsEcF1QVRPhJMcJZ+1GtH
mX2JaSKciKJRfoEMkO7bO2MJ+00qB1z7rnO.8.5uPqV800qVYmxLwpwqWErm
+71JducYA6E887GEAgtOSIdVYmW0rYX2NQGxM79cC0HLphuQTYeO4vcTOrWV
WIT1uN5qS1LnNLKYb1VtZ6+Hqe7ichsJ2lLIFGgCQoIlWIYLy.kFgQ+83839
ZTeoU3tgffW9HYocm1r4SumFXl5aqWadI7Mf5D3odJiXXadwxndLjTmgiPDb
DLDjfowFRwxrZLMdtTjTLGNVK9r9l+IJtqZPVFokVZkrWg9PeyP2VwwZ8G5U
7NMa3fI0D5nIf3FR7vRX1cSaMxfTwoDshCjfWTjY0aVwB06z25v4ixjATdLh
k5DpKQV5roNaNTeyfR0LMSH2dTKl43.wohGd87LYCu9woYCN3Ef536EJQ2GE
07MU16DeGCpgMi1XbLIKeAwwY.lT6OQTLTopJhnIFEmZikoWmgx5oIHCxn31
NCLfPw73TKoxI9xvoctiAT.cdpe+ZdrSGMSu.WYNWThcPGaecxtqpgqszWJr
lBd8Xa53dg56WIw7bM.LrKwrWpc34TbxcrMSfZxD6D93hkUWJGbEujq3VAuT
tUIap4ceANU+.sKrQ5jjzEn5ypYDCohdRJ97MvG744CU2lVlNqbbwXsYZTt4
ZRVxQb5jJuPpXc5co.UNru8F7JodPGm+XV7RBJwPlGVgjl+tAUicwVohLkRc
4dhw9UJM7jgooLCPJ2aKstCIpU5Pv2Ie.8No9+IMArYzMo0D3N+yX6wyoXKF
xZNC+mDQtkFltfHGO9nKrDzmi7XKqF9534T4LH8qs4iZ4a+WjIu6CnsMU8OP
gQiyn1fY1nkfRttM30b9SScBzT+QcJrVXNfzA55hpSSWNaw+pOymS4JRV7A9
RRuqG36MsGBRbbTxKQ1jz7ETjJA1ZTaPaqD7NX34XXKyMnOnPhuYvbahI4Kn
OPV8ARJORhQ+AM7VaFwi34bmAHAO2GDK62qdjG4cLC.+6SxWaWPPkr9G+kHr
aBy7mZObOrlQHbG1HD85tnTzqj0bC+OZMot0LoE2WbzUX0D+J.oyNCDRzqfj
osN.fh3g0KFDjvdfzgsyxPxCfHoNwbQ.kj5AR5jaKmRG9RtrwKFJjtlWdBDH
cvq5x5DD.cmRP3EevT.b7n2IFQ7JTp..FQ7QjHP3MPX2KiG1KNgyAAKlGgsL
vPxO2bXvh5EVzEikWoXOsH1dYYai9D4ictPF+AacOGwbp6mFwb0h2b9r2NYM
MckhNamdu8HSNCz3kAsGdzuDi8SMTxaaeRz0OtbKH5Vo+TicikGZuTV6tL0d
Ym3I4g0mXmg2o6BVoaAdnytwBdNMNvcqMZJVOHGMvZ5ogz1lds9br8sbGSrc
yu9aq+efewhiV
</code></pre>

balintlaczko:
Thanks!! I am on it — but I won't be able to implement anything before Thursday, unfortunately. I'll try to think of something fun! :))

tremblap:
no worries no stress FluCoMa volunteer coding time is definitely not under pressure :D

tremblap:
and very appreciated

weefuzzy (approved):

PR #306, opened 2025-06-10
optimise spikes to times sorting

jamesb93:
This fix makes the sorting in this part of spikesToTimes much faster, for reasons...


tremblap (approved):
this yields the same results in all UTs

PR #307, opened 2025-06-10
sort windows version pre-deprecation

tremblap:
changed the one instance...


jamesb93:
This looks good to me. I think we can just merge.

PR #309, opened 2025-06-17
Fix/trigger output logic

tremblap:
This fixes the logic of the trigger output of the KR objects.

Instead of copying the input to the output, we now output 0 except in the block where the kr action will have been successful. If the process function bails, it will give the user nothing, which is a good way to troubleshoot and in line with trigger chaining.


tremblap:
thanks it's merged and now the nightlies have them - feel free to road test!

tedmoore:
will do but after my gigs this week

PR #311, opened 2025-06-20
FluidDataSet: `point` args in `add` and `update` are now const

### LINKED ISSUE #80: `FluidDataSet`: can `point` args in `add` and  `update` be made `const`?
Looks like it shouldn't be problematic, and would make consistent with constructors ﻿

tremblap:
road tested on all UTs, fixes #80

### DIFF
--- a/include/flucoma/data/FluidDataSet.hpp
+++ b/include/flucoma/data/FluidDataSet.hpp
@@ -104,7 +104,7 @@ bool add(idType const& id, FluidTensorView<const dataType, N> point)
       return pos->second;
   }
 
-  bool update(idType const& id, FluidTensorView<dataType, N> const point)
+bool update(idType const& id, FluidTensorView<const dataType, N> point)
   {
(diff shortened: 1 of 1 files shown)


include/flucoma/data/FluidDataSet.hpp
```diff
@@ -104,7 +104,7 @@ class FluidDataSet
       return pos->second;
   }
 
-  bool update(idType const& id, FluidTensorView<dataType, N> point)
+  bool update(idType const& id, FluidTensorView<dataType, N> const point)
```
weefuzzy, line 107: ditto 
```suggestion 
bool update(idType const& id, FluidTensorView<const dataType, N> point)
```

tremblap:
ok it all passes uts so i'd merge

tremblap:
thanks for the explanation. The templating understanding of yours truly is still fragile...

PR #312, opened 2025-06-20
dataset and kdtree - k nearest is checked for min number of k

### LINKED ISSUE #184: `KDTree`: Assertion failure with `numNeighbours` < 0
It is possible to set the `numNeighbours` control of `KDTreeClient` to < 0. However, the algorithm itself explicitly expects this to be nonnegative, and calls `asUnsigned`. Presumably in release builds, this will instead wrap (badness). We should enforce this invariant in the interface by adding `Min(0)`.

tremblap:
fixes #184 by adding checks both in the parameter description and at the recup of optional arguments

### DIFF
--- a/include/flucoma/clients/nrt/DataSetClient.hpp
+++ b/include/flucoma/clients/nrt/DataSetClient.hpp
@@ -350,7 +350,7 @@ namespace dataset {
 
 constexpr auto DataSetReadParams = defineParameters(
     InputDataSetClientRef::makeParam("dataSet", "DataSet Name"),
-    LongParam("numNeighbours", "Number of Nearest Neighbours", 1),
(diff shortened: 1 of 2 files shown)


tremblap:
in the meantime, the fact that the optional can go negative was crashing, so I fixed both to be consistent for now

tremblap:
@weefuzzy if you are happy with my answer I'll merge this (as it fixes crashes) and add a feature request to align the Knearest between the 2 (adding 0 and range to ds.knearest() which will be good for my stl iterator learning)

weefuzzy:
> because 0 in kdtree is 'give me all that respect the distance' and we didn't implement that in DS - which could be its own ticket if we decide to.

Right, ok, so to do with radius search for kdtree? Fine.

PR #314, opened 2025-06-25
Update flucoma-core-ci.yml

weefuzzy:
Weird windows test fails correlated with switching CI to `windows-latest`, seeing if rolling back fixes this or if cause is something esle


weefuzzy:
Nope, still barfs

